PR #16, opened 2014-04-18
type-o

h4ck3rm1k3:
fixing a typo


haosdent:
This has been fixed in latest code https://github.com/apache/mesos/blob/1.1.x/include/mesos/zookeeper/zookeeper.hpp#L44 We should close this pr.

PR #23, opened 2014-07-24
Clean help flags

kensipe:
in the case of the zk flag.  the help is wrong.  The full file:/// protocol **must** be followed.  the way it is written, it appears as though you can just provide the absolute path, which fails.

The rest of it, cleans up the help to be more consistent with the phrasing.


jfarrell:

PR #28, opened 2014-10-29
Update mpiexec-mesos.in

strat0sphere:
In the if statement [[ is used - [[ is a bash builtin command and is not available in sh. It will give an error ( [[: not found ) unless /bin/sh is a symbolic link to /bin/bash

### DIFF
--- a/mpi/mpiexec-mesos.in
+++ b/mpi/mpiexec-mesos.in
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 #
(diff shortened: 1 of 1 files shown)


kaysoky:
This MPI code likely does not work anymore, as there is no maintainer for it.  I've opened https://issues.apache.org/jira/browse/MESOS-6084 to track the deprecation of the MPI framework.

jfarrell:
Closing as requested in https://s.apache.org/V8r3

PR #34, opened 2015-01-19
Replace hardcoded 'http:' in webui code with window.location.protocol

arnaudbriche:
Hi,

In my mesos setup I access mesos webui from on https endpoint which reverse proxy requests to mesos (listening on the private interface ip of the machine); in order to not break internal links in the webui, 
I just replaced the hardcoded 'http:' string with a protocol independent window.location.protocol.
Maybe it can be of interest for some other mesos users.


bmahler:
Thanks Arnaud! Nice, there will be built-in HTTPS support in Mesos at some point, you may want to chime in here:

https://issues.apache.org/jira/browse/MESOS-1825

akerekes:
Is there a plan to include this PR in the meantime before actual HTTPS support is implemented?

akerekes:
@arnaudbriche There is one more location where this needs to be fixed:
https://github.com/apache/mesos/blob/master/src/webui/master/static/browse.html#L71
It should be:
```
href="//{{slave_host}}/files/download.json?path={{encodeURIComponent(file.path)}}">
```

arnaudbriche:
Thanks @akerekes for pointing this out.
It should also be noted that when one tries to connect to a non-leader node, mesos (server-side) issue a 301 Redirect with an hardcoded http:// protocol. 
I did took a quick look at the C++ code,  but didn't find an easy way to fix it though.

jfarrell:

PR #36, opened 2015-02-23
replace unsafe "find | xargs" with "find -exec"

lloesche:
The current `find | xargs rm -rf` in the Makefile could potentially destroy data if mesos source was in a folder with a space in the name. E.g. if you for some reason checkout mesos to `/ mesos` the command in src/Makefile.am would turn into a `rm -rf /`

`find | xargs` should be NUL delimited with `find -print0 | xargs -0` for safer execution or can just be replaced with the find build-in option `find -exec '{}' \+` which behaves similar to xargs.

Example:
```
$ touch somefile.txt
$ touch "file with spaces.txt"
$ find . -name "*.txt" | xargs -n 1 echo rm -rf
rm -rf ./file
rm -rf with
rm -rf spaces.txt
rm -rf ./somefile.txt
$ find . -name "*.txt" -exec echo rm -rf '{}' \;
rm -rf ./file with spaces.txt
rm -rf ./somefile.txt
```
(Note: for demonstration purposes `xargs -n 1` is similar to `find -exec '{}' \;` whereas `xargs` default behaviour is similar to `find -exec '{}'  \+`)

There was a second occurrence of this in a test script, though in that case it would only rmdir empty folders, so is less critical.

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -1191,7 +1191,7 @@ endif # HAS_PYTHON
 
 clean-python:
 	find python -name "build" -o -name "dist" -o -name "*.pyc"		\
-		-o -name "*.egg-info" | xargs rm -rf
+		-o -name "*.egg-info" -exec rm -rf '{}' \+
 
 PHONY_TARGETS += clean-python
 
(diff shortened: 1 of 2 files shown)


vinodkone:
I think this broke the ReviewBot. See https://issues.apache.org/jira/browse/MESOS-5958

PR #46, opened 2015-06-17
fixed broken 404 doc link for torque on mesos

kensipe:
It is unclear if Torque is an active project... I can remove the torque docs if it is preferred.   However the doc links for torque resulted in a 404 and this PR fixes that issue with a relevant link.


adam-mesos:
Thanks @kensipe but the Torque integration project is inactive, unsupported, and may no longer work. We've recently discussed moving it out of the Mesos repo into a separate project, if somebody wants to own/maintain it. So, rather than linking to something outdated (or 404), we should just remove the link.
See https://issues.apache.org/jira/browse/MESOS-2311 and https://reviews.apache.org/r/31595/ for more context.

jfarrell:

PR #49, opened 2015-06-30
MESOS-2424 FIX: protobuf installs latest setuptools in Mesos dist if …

kamilchm:
…there's older one in build system OS https://github.com/google/protobuf/blob/v2.5.0/python/setup.py#L187


jfarrell:

PR #51, opened 2015-07-29
[MESOS-3170] Add $LIBS to build path of the CRAM-MD5 test

hellertime:
If the sasl2 library has been statically linked than the CRAM-MD5 check in the configure script can fail due to missing symbols. The correct approach is to add the required static lib archives into the LIBS variable. Adding this variable to the build action will allow the crammd5_installed.c check program compile and allow the check to pass.


jfarrell:

PR #59, opened 2015-08-25
fix missing timeout in mesos-fetcher

T0aD:
Added hardcoded timeout in mesos-fetcher code, following lots of issues with hanging connections, that would forbid the executor to proceed.

Notes:
- code tested on 0.22.0 (slighty different, just one place where the patch resides)
- duplicated hardcoded values (bad bad bad), make it configurable somehow ?
- code non compiled / tested, since MESOS_FETCHER_INFO variable content changes and I wasnt able to start a download with the old json string (despite adding sandbox_directory).


kaysoky:
Dealing with large artifacts should be the framework's responsibility.  

Suppose the framework launches a task with a URI that sometimes hangs.  The framework will receive a `TASK_RUNNING` status update as soon as the URI is done fetching.  If this status update does not come without some framework-defined timeout, the framework should then remediate, say by killing the task and trying again.

We don't want Mesos to dictate the maximum timeout for any fetches.

jfarrell:
Closing as requested in https://s.apache.org/V8r3

PR #63, opened 2015-09-02
adding new chef cookbook

acaiafa:
Adding a new more reusable cookbook for setting up mesos. It supports all of the options out of the box.


acaiafa:
ping!

PR #70, opened 2015-10-09
Documentation file naming cleanup

elsmorian:
Part of https://issues.apache.org/jira/browse/MESOS-3686

Clean up of filenames in documentation, prefer hyphens and drop 'mesos' from filenames


elsmorian:
@davelester

PR #71, opened 2015-10-09
Use character entity references instead of angle brackets

elsmorian:
Fixes https://issues.apache.org/jira/browse/MESOS-2237


elsmorian:
@davelester

PR #79, opened 2015-10-27
Patch Docker containerizer for container name

cfortier2:
Use 'slaveId.containerId' for Docker slave recovery. This should close https://issues.apache.org/jira/browse/MESOS-3808

@tnachen It looks like you worked on this bit of code last. Please provide any thoughts.


tnachen:
Technically this is not correct as the container ID shouldn't have the SlaveID in there and will cause other problems. I'm not sure what the recovery problem as I haven't looked into yet, but I know this is not how we should fix this. We should fix the logic that's happening in recover instead.

jfarrell:

PR #83, opened 2015-12-14
Added Qunar to the powered by mesos list.

BrickXu:
We use mesos and marathon to support the log analyize programs. It is a great resource mananger to hold thousands of applications to deal with 6~8 billion lines text per day, thanks very much!


artemharutyunyan:
Hello BrickXu, thanks for the PR. Mesos project is actually using Review Board for accepting patches. I went ahead and created a patch with your change (https://reviews.apache.org/r/41706/), it should be pushed to Mesos shortly.

BrickXu:
@artemharutyunyan Thank you :)

PR #90, opened 2016-02-02
[MESOS-4370] Retrieve network mode and use that to get that networks IP address.

travishegner:
This patch will first query the docker API for the HostConfig.NetworkMode, which is populated with the network name. (Essentially what was passed in --net <name> to the docker run command). This name is then used as a key in NetworkSettings.Networks.<name>.IPAddress to get the IP address that is currently in use by the container.

It appears that even though the docker API has been set up to allow for multiple networks, our testing has indicated that it's still only applying one network to the container (the last one via the --net argument on the run line). I can only speculate that the docker API will change again in the near future, but I can't speculate how, so at least this fixes the problem as it stands right now. This patch is tested against Docker 1.9.1 on Ubuntu 14.04.

Continues #87

### DIFF
--- a/src/docker/docker.cpp
+++ b/src/docker/docker.cpp
@@ -301,14 +301,36 @@ Try<Docker::Container> Docker::Container::create(const string& output)
 
   bool started = startedAtValue.get().value != "0001-01-01T00:00:00Z";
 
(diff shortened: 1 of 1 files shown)


jfarrell:

PR #93, opened 2016-03-28
Upgrade zookeeper to 3.4.8.

chenzhiwei:


3rdparty/Makefile.am
```diff
@@ -51,7 +51,7 @@ EXTRA_DIST =		\
 EXTRA_DIST +=		\
   $(LEVELDB).patch
 
-# We need to patch ZooKeeper in order to get 3.4.5 to compile on
+# We need to patch ZooKeeper in order to get 3.4.8 to compile on
 # OS X 10.10. See: MESOS-1797.
```
vinodkone: Does this comment still apply? AFAICT MESOS-1797 doesn't apply to ZK 3.4.8 since that issue has been resolved in ZK 3.4.7. Looks all the patch does is apply PPC specific fixes for which you already submitted a review https://reviews.apache.org/r/45376/.

I would recommend to kill the zookeper.patch stuff in this review altogether since 3.4.8 doesn't need any patches per se. Then in https://reviews.apache.org/r/45376/ you can add back the patch stuff and mention that you need the patch to compile it on PPC.

chenzhiwei: Thank your for pointing this, I will remove this comment. And for the zookeeper patch file I think I already removed it.

chenzhiwei:
I found I need to rebase again if the upgrade libev patch was submitted, so let's wait for libev patch to be submitted, then I will rebase.

vinodkone:
libev is committed. You can rebase it now.

PR #99, opened 2016-04-27
Increased status timeout for LinuxFilesystemIsolatorTest.ROOT_Recover…

Djidiouf:
…OrphanedPersistentVolume

Fix the following issue when doing make check on RedHat7:

../../src/tests/containerizer/filesystem_isolator_tests.cpp:697: Failure
Failed to wait 15secs for status
../../src/tests/containerizer/filesystem_isolator_tests.cpp:685: Failure
Actual function call count doesn't match EXPECT_CALL(sched, statusUpdate(&driver, _))...
         Expected: to be called at least once
           Actual: never called - unsatisfied and active

### DIFF
--- a/src/tests/containerizer/filesystem_isolator_tests.cpp
+++ b/src/tests/containerizer/filesystem_isolator_tests.cpp
@@ -702,7 +702,7 @@ TEST_F(LinuxFilesystemIsolatorTest, ROOT_RecoverOrphanedPersistentVolume)
       {offer.id()},
       {CREATE(persistentVolume), LAUNCH({task})});
 
-  AWAIT_READY(status);
+  AWAIT_READY_FOR(status, Seconds(60));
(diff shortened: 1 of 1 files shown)


Gilbert88:
Hi @Djidiouf , thanks for this PR. This is a known issue. Currently we have some tests in LinuxFilesystemIsoaltorTest are flaky. We used to trying fix them by increasing the await time from 15 sec to 60 sec, but it the tests are still flaky, because that may not be the root cause for this failure.

We usually track on mesos issues in apache mesos JIRA (https://issues.apache.org/jira/browse/MESOS/), which makes it easier to classify diff types and priority of  any issue. If you are interested in working on any issue, please follow [this](http://mesos.apache.org/documentation/latest/submitting-a-patch/) document and submit patches to apache mesos reviewboard. :)

jfarrell:

PR #118, opened 2016-06-30
Make Mesos healtchecks compatible with Marathon

brat002:
https://github.com/mesosphere/marathon/issues/916

### DIFF
--- a/src/health-check/main.cpp
+++ b/src/health-check/main.cpp
@@ -132,16 +132,14 @@ private:
   {
     VLOG(1) << "Check passed";
 
-    // Send a healthy status update on the first success,
-    // and on the first success following failure(s).
-    if (initializing || consecutiveFailures > 0) {
-      TaskHealthStatus taskHealthStatus;
-      taskHealthStatus.set_healthy(true);
-      taskHealthStatus.mutable_task_id()->CopyFrom(taskID);
-      send(executor, taskHealthStatus);
-      initializing = false;
-    }
-    consecutiveFailures = 0;
+    // Send a healthy status update on every check, because https://github.com/mesosphere/marathon/issues/916
+    
+    TaskHealthStatus taskHealthStatus;
+    taskHealthStatus.set_healthy(true);
+    taskHealthStatus.mutable_task_id()->CopyFrom(taskID);
+    send(executor, taskHealthStatus);
+    initializing = false;
+    
     reschedule();
   }
 


src/health-check/main.cpp
```diff
@@ -132,16 +132,14 @@ class HealthCheckerProcess : public ProtobufProcess<HealthCheckerProcess>
   {
     VLOG(1) << "Check passed";
 
-    // Send a healthy status update on the first success,
-    // and on the first success following failure(s).
-    if (initializing || consecutiveFailures > 0) {
-      TaskHealthStatus taskHealthStatus;
```
haosdent: Hi, why remove this condition `if (initializing || consecutiveFailures > 0) {`

brat002: We need to update Health status every time when it triggered.

haosdent: Oh, got it. Could you follow https://github.com/apache/mesos/blob/master/docs/submitting-a-patch.md to submit the patch at https://reviews.apache.org/groups/mesos/ 
We use reviewboard to review code changes.

brat002: Could anybody make correct pull request? I am not C programmer and not 100% sure in these changes.

haosdent: OK, LGTM. If you are not familiar with review board, I could help you to post this patch there.

brat002: It will be great. Thank you.

drexin: Could we please make this behavior optional? In a cluster with lots of tasks this would create a lot of unnecessary traffic.

ConnorDoyle: +1 to @drexin's comment; cutting down communication between the task host and the scheduler was the original motivation behind implementing health checks in the Mesos agent :)

Other options: 1) let the framework persist task health status before acknowledging the update 2) have a deeper look at the status updates that are synthesized by the master during reconciliation, maybe trigger the executor to send the current health?

haosdent: >1) let the framework persist task health status before acknowledging the update
This approach looks better. @brat002 May you mind close this pull request and let me create a ticket in https://issues.apache.org/jira/browse/MESOS to track this?

bbannier:
Could you please move this to reviewboard so we got this growing discussion in the expected place? @drexin @haosdent @brat002 @ConnorDoyle

brat002:
>May you mind close this pull request and let me
Yes, of course.

2016-07-07 9:04 GMT+03:00 Benjamin Bannier <notifications@github.com>:

haosdent:
Thanks @bbannier I file the ticket at https://issues.apache.org/jira/browse/MESOS-5803 , let's continue the discussion there.

janisz:
@brat002 I think this should be closed as it turns out it was a bug in Marathon.

gkleiman:
Closing this PR as the Marathon bug has been fixed and we don't want to introduce this behaviour to Mesos.

PR #130, opened 2016-07-07
Avoid 'Permission denied' error with 'sudo cat >'

surdy:
Not sure if this minor change warrants creation of a JIRA and a review board request. I will do so if required. 

`sudo cat > /etc/yum.repos.d/wandisco-svn.repo << EOF` leads to `bash: /etc/yum.repos.d/wandisco-svn.repo: Permission denied` error. 

`sudo bash -c 'cat > /etc/yum.repos.d/wandisco-svn.repo <<EOF` fixes the problem.

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -98,14 +98,14 @@ Following are the instructions for stock CentOS 6.6. If you are using a differen
     # 'Mesos > 0.21.0' requires 'subversion > 1.8' devel package, which is
     # not available in the default repositories.
     # Create a WANdisco SVN repo file to install the correct version:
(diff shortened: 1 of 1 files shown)


klueska:
Looks good to me.  To summarize the issue that he's fixing a bit more clearly. Just running:
```
sudo cat > /etc/yum.repos.d/wandisco-svn.repo
```
Doesn't actually do what's intended because anything after the redirect (i.e. the `>`) isn't picked up by the sudo.  We need to run `bash -c "<argument>"` instead so that sudo launches a subshell and the redirect happens inside the subshell.

PR #133, opened 2016-07-08
Output disk resource source information.

timcharper:
This information is obscured when formatted via `stringify`, leading to confusing error messages such as:

```
Task uses more resources
cpus(*):4; mem(*):4096;     ports(*):[31000-31000]; disk(kafka, kafka)[kafka_0:data]:960679
than available
cpus(*):32; mem(*):256819;  ports(*):[31000-32000]; disk(kafka, kafka)[kafka_0:data]:960679;   disk(*):240169;
```

(above was formatted and aligned for clarity)

The validation error here is actually complaining that the disk requested didn't have any source information (and therefore defaulted to a root volume), but the available persistent volume was a mount source.

### DIFF
--- a/src/common/resources.cpp
+++ b/src/common/resources.cpp
@@ -1475,6 +1475,19 @@ ostream& operator<<(ostream& stream, const Volume& volume)
   return stream;
 }
 
+ostream& operator<<(ostream& stream, const Resource::DiskInfo::Source& source)
+{
+  switch (source.type()) {
+  case Resource::DiskInfo::Source::MOUNT:
+    stream << ":mount:" << source.mount().root();
+    break;
+  case Resource::DiskInfo::Source::PATH:
+    stream << ":path:" << source.path().root();
+    break;
+  }
+
+  return stream;
+}
 
 ostream& operator<<(ostream& stream, const Resource::DiskInfo& disk)
 {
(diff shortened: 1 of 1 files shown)


src/common/resources.cpp
```diff
@@ -1475,6 +1475,20 @@ ostream& operator<<(ostream& stream, const Volume& volume)
   return stream;
 }
 
+ostream& operator<<(ostream& stream, const Resource::DiskInfo::Source& source)
+{
+  stream << source.type();
+
+  if (source.has_path()) {
```
rukletsov: It looks like the common pattern is to switch on the type. Could you please follow this pattern? You can convert `has_*()` into `CHECK`s.

timcharper:
Not sure why the tests have failed, but @rukletsov - I've updated the PR; it doesn't seem like a good idea to put a `CHECK` call into a string formatter, so I didn't. Do you disagree?

timcharper:
I ran the tests manually and they still pass

kaysoky:
The travis CI is broken/not-maintained, since we don't (usually) make any code changes via GitHub PRs.  See: http://mesos.apache.org/documentation/latest/submitting-a-patch/

Can you file a JIRA for this change?  We might have to consider backwards compatibility if the resources string changes.  And we'd like that discussion to take place in JIRA.

rukletsov:
Why would we want to back port such change?

timcharper:
I've followed the instructions and the JIRA issue is here: https://issues.apache.org/jira/browse/MESOS-5824

rukletsov:
Could you please close this PR since there is a review instead now?

PR #141, opened 2016-07-15
Fix FrameworkInfo::user documentation

ktf:
Add a note about the user being required in case the HTTP Scheduler API
is used.

### DIFF
--- a/include/mesos/mesos.proto
+++ b/include/mesos/mesos.proto
@@ -205,9 +205,14 @@ message MachineInfo {
  * Describes a framework.
  */
 message FrameworkInfo {
-  // Used to determine the Unix user that an executor or task should
-  // be launched as. If the user field is set to an empty string Mesos
-  // will automagically set it to the current user.
+  // Used to determine the Unix user that an executor or task should be
+  // launched as.
+  //
+  // When using the MesosSchedulerDriver, if the field is set to an
+  // empty string, it will automagically set it to the current user.
+  //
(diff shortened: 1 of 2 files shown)


haosdent:
Please update `include/mesos/mesos.proto` as well.

ktf:
Done.

haosdent:
LGTM. :-)

include/mesos/mesos.proto
```diff
@@ -205,9 +205,12 @@ message MachineInfo {
...
 message FrameworkInfo {
-  // Used to determine the Unix user that an executor or task should
-  // be launched as. If the user field is set to an empty string Mesos
-  // will automagically set it to the current user.
+  // Used to determine the Unix user that an executor or task should be
+  // launched as. If the user field is set to an empty string
+  // MesosSchedulerDriver will automagically set it to the current user.
+  //
+  // NOTE: The user has to be set explicitly in case the HTTP API
+  // is used.
```
zhitaoli: Can we restructure these two sentences as this is not very clear.

Maybe consider something among:

```
// Used to determine the Unix user that an executor or task should be
// launched as.
//
// If using HTTP API, this field must be explicitly set.
//
// If using MesosSchedulerDriver and this field is set to an empty string, MesosSchedulerDriver
// will automagically set it to the current user.

ktf:
Hopefully improved.

Gilbert88:
Thanks for the clarification @ktf .

This patch LGTM!

PR #152, opened 2016-07-28
Add Mesos books to documentation page

rji:
This commit adds the known Mesos books to-date to the Mesos documentation page so that community members can easily find them.

Testing: built the site.

screen shot 2016-07-28 at 12 40 49 pm


rji:
Would someone mind building and pushing the website? It looks like there isn't a CI job to do this.

PR #153, opened 2016-07-29
[MESOS-5925] Building Mesos in Docker Beta on OS X fails - tar issue …

radekg:
…in Makefile

### DIFF
--- a/3rdparty/Makefile.am
+++ b/3rdparty/Makefile.am
@@ -129,7 +129,7 @@ CLEAN_EXTRACTED =	\
 # which cause the packages to get extracted as necessary. We also
 # apply any patches as appropriate.
 %-stamp: %.tar.gz
-	gzip -d -c $^ | tar xf -
+	gzip -d -c $^ | tar xf - --no-same-owner
 	test ! -e $(srcdir)/$*.patch || patch -d $* -p1 <$(srcdir)/$*.patch
 	touch $@
 
--- a/3rdparty/libprocess/3rdparty/Makefile.am
+++ b/3rdparty/libprocess/3rdparty/Makefile.am
@@ -81,7 +81,7 @@ CLEAN_EXTRACTED =	\
 # which cause the packages to get extracted as necessary. We also
 # apply any patches as appropriate.
 %-stamp: $(BUNDLED_DIR)/%.tar.gz
-	gzip -d -c $^ | tar xf -
+	gzip -d -c $^ | tar xf - --no-same-owner
 	test ! -e $(top_srcdir)/../$*.patch || patch -d $* -p1 <$(top_srcdir)/../$*.patch
 	touch $@
 


haosdent:
Run `man tar` in OS X could not find this option: `no-same-owner`.  I think we should make sure this patch don't break OS X?

haosdent:
Verify OS X tar would ignore unknow flags. So it works in OS X. 

+1 for this change.

tillt:
Why exactly do we need this patch?

I understand it solves the problem for the OP but WHY does the problem even occur?
As far as I understand, the option `--no-same-owner` is only effective when being invoked by `root` - so if `root` was the current user, why would we get a permission denied? I would like to understand the actual problem.

haosdent:
It looks like a bug of docker mount volume in OS X.

```
root@7f689b81a4bb:~/workspace/cpp/mesos/build/gmock-1.7.0# chown root:root test/gmock_test_utils.py
chown: changing ownership of 'test/gmock_test_utils.py': Permission denied
```

tillt:
If it truly is a bug of docker, then we should get that fixed upstream, no?

haosdent:
@tillt That make sense. I am finding if it has tracked in docker issues, otherwise I could file one.

radekg:
@haosdent:

```
[rad] Downloads $ uname -a
Darwin moan.fttx-di01-b07.german-fiber.net 15.4.0 Darwin Kernel Version 15.4.0: Fri Feb 26 22:08:05 PST 2016; root:xnu-3248.40.184~3/RELEASE_X86_64 x86_64
[rad] Downloads $ ls -la | grep zookeeper
-rw-r--r--@   1 rad  staff    22261552 Jun  7 16:57 zookeeper-3.4.8.tar.gz
[rad] Downloads $ tar xvf zookeeper-3.4.8.tar.gz --no-same-owner
x zookeeper-3.4.8/
x zookeeper-3.4.8/zookeeper-3.4.8.jar.md5
x zookeeper-3.4.8/zookeeper-3.4.8.jar.asc
x zookeeper-3.4.8/zookeeper-3.4.8.jar
x zookeeper-3.4.8/contrib/
x zookeeper-3.4.8/contrib/ZooInspector/
x zookeeper-3.4.8/contrib/ZooInspector/icons/
x zookeeper-3.4.8/contrib/ZooInspector/icons/launch_run.gif
x zookeeper-3.4.8/contrib/ZooInspector/icons/save_edit.gif
x zookeeper-3.4.8/contrib/ZooInspector/icons/new_con.gif
x zookeeper-3.4.8/contrib/ZooInspector/icons/search_prev.gif
x zookeeper-3.4.8/contrib/ZooInspector/icons/file_obj.gif
x zookeeper-3.4.8/contrib/ZooInspector/icons/launch_stop.gif
x zookeeper-3.4.8/contrib/ZooInspector/icons/jspdecl.gif
x zookeeper-3.4.8/contrib/ZooInspector/icons/fldr_obj.gif
x zookeeper-3.4.8/contrib/ZooInspector/icons/refresh.gif
x zookeeper-3.4.8/contrib/ZooInspector/icons/info_obj.gif
x zookeeper-3.4.8/contrib/ZooInspector/icons/search_next.gif
...
[rad] Downloads $ ls -la | grep zookeeper
drwxr-xr-x@  22 rad  staff         748 Feb  6  2016 zookeeper-3.4.8
-rw-r--r--@   1 rad  staff    22261552 Jun  7 16:57 zookeeper-3.4.8.tar.gz
```

@tillt I'm all for fixing this in docker, the only thing that worries me is the amount of time it takes docker to fix this (if this is a bug and not a _feature_ of the new Docker for mac). The `--no-same-owner` option has zero impact on anything and works fine in every case. The patch does not break anything. It seems that the tar option was designed specifically for such situations, why not to take advantage of it and have the `Makefile` working in every situation and report to docker that the problem exists. Thoughts?

haosdent:
As I checked, only gmock-1.7.0 and gperftools-2.5 encounter this problem. If I use the releases from https://github.com/google/googlemock/releases, it works fine.

And for other tars, although there are packaged with different users. They still could be unpacked successfully.

```
$ tar --list -v --file boost-1.53.0.tar.gz  | head -2
drwxr-xr-x jie/staff         0 2016-01-18 07:35 boost-1.53.0/
drwxr-xr-x jie/staff         0 2016-01-18 07:34 boost-1.53.0/boost/
```

haosdent:
OK, I think I find the root cause. 

```
$ stat test/gmock-port_test.cc
  File: 'test/gmock-port_test.cc'
  Size: 2020            Blocks: 4          IO Block: 512    regular file
Device: 28h/40d Inode: 37331162    Links: 1
Access: (0444/-r--r--r--)  Uid: (    0/    root)   Gid: (    0/    root)
Access: 1970-01-01 00:00:00.000000000 +0000
Modify: 2013-09-18 17:50:15.000000000 +0000
Change: 2016-08-14 04:06:55.000000000 +0000
 Birth: -
```

```
$ stat ./.git/objects/pack/pack-4ad0e5887c21e2edcca6e6b0af3851568bbee729.idx
  File: './.git/objects/pack/pack-4ad0e5887c21e2edcca6e6b0af3851568bbee729.idx'
  Size: 144180          Blocks: 282        IO Block: 512    regular file
Device: 28h/40d Inode: 37330833    Links: 1
Access: (0444/-r--r--r--)  Uid: (    0/    root)   Gid: (    0/    root)
Access: 2016-08-14 04:08:46.000000000 +0000
Modify: 2016-04-19 21:45:36.000000000 +0000
Change: 2016-08-14 04:06:51.000000000 +0000
 Birth: -
```

Those files are `-r--r--r--`, which don't allow to write. 😂 Docker is correct.

radekg:
Fix in https://github.com/apache/mesos/pull/158

PR #157, opened 2016-08-13
[MESOS-6038] Prevent memory leaks

aaronjwood:
This should prevent any of the promises that are created in the various `ZookeeperProcess` class methods from leaking memory.

https://reviews.apache.org/r/51068/

### DIFF
--- a/docs/contributors.yaml
+++ b/docs/contributors.yaml
@@ -1,3 +1,13 @@
+- name: Aaron Wood
+  affiliations:
+    - {organization: Verizon Labs}
+  emails:
+    - aaronjwood@gmail.com
+    - me@aaronjwood.com
+    - aaron.wood@verizon.com
+  jira_user: aaron.wood
+  reviewboard_user: aaron.wood
+
 - name: Adam B
   affiliations:
     - {organization: Mesosphere}
--- a/src/zookeeper/zookeeper.cpp
+++ b/src/zookeeper/zookeeper.cpp
@@ -344,23 +344,20 @@ public:
       bool watch,
       vector<string>* results)
   {
-    Promise<int>* promise = new Promise<int>();
-
-    Future<int> future = promise->future();
+    Promise<int> promise = Promise<int>();
 
     tuple<Promise<int>*, vector<string>*>* args =
-      new tuple<Promise<int>*, vector<string>*>(promise, results);
+      new tuple<Promise<int>*, vector<string>*>(&promise, results);
 
     int ret =
       zoo_aget_children(zh, path.c_str(), watch, stringsCompletion, args);
 
     if (ret != ZOK) {
-      delete promise;
       delete args;
       return ret;
     }
 
-    return future;
+    return promise.future();
   }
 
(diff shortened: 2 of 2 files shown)


haosdent:
Hi, @aaronjwood Because we use [reviewboard](https://reviews.apache.org/groups/mesos/) to review code changes. Could you follow https://github.com/apache/mesos/blob/master/docs/submitting-a-patch.md to submit your patch?

haosdent:
In addition, I think it would be more helpful to describe more context about the memory leaks here to help others understand this patch.

xiaods:
could you please describe the context about memory leaks and give some testing to confirm the root cause.

aaronjwood:
Hi @haosdent. Sure, I'm sorry. I had created a JIRA and reviewboard account and went to open this PR but should have only included the change to the contributors file.

aaronjwood:
For the context: I had found a few places where these `Promise` objects were leaking. Once the `Future` was returned there was no way to free the memory for the `Promise` object that had been allocated thus causing the leak. I don't have any direct issues with running Mesos that stemmed from this, it's just something that I thought could be fixed.

Vish0007:
@aaronjwood  the "promise" object you have has local scope (and lifetime) and its not valid outside the method; I am sure valgrind wold report this as error;

haosdent:
+1 For what @Vish0007 said. If promise became a locale scope object here, return its future seems incorrect. I search `return \S+\.future\(\);` in all exists code, all `Promise` are member objects or heap objects.

aaronjwood:
I might be misunderstanding things but why can we not do this? I see it being done here https://github.com/apache/mesos/blob/master/3rdparty/libprocess/src/http.cpp#L1001

If we make the `Promise` local like it is in this patch, pass its address into the tuple via the copy constructor (that should be the copy constructor getting called, right?) so that it takes a copy and doesn't rely on an actual address from the stack, and return the future in the same way as the example above, would that not work? Even though the `Promise` is local to the stack, when we return `Future` shouldn't it be a copy of what's needed underneath?

haosdent:
>Even though the Promise is local to the stack, when we return Future shouldn't it be a copy of what's needed underneath?

Sorry, I misunderstanding something before. Future store its data in `std::shared_ptr<Data> data` and your function return a copy instead of const reference, so it should work.

aaronjwood:
No no, that's okay. I appreciate everyone helping me validate these changes. I'm trying to be very careful with this :)

haosdent:
By the way, copy a comment from @jpeach in https://mesos.slack.com/messages/cxx/

```
Promise<int> promise = Promise<int>();

why not just
```Promise<int> promise;```
``

aaronjwood:
I don't have access to their Slack but I agree.

haosdent:
You could join via https://mesos-slackin.herokuapp.com/

aaronjwood:
It looks like I was wrong about this, I had thought the copy constructor of tuple was getting called when it wasn't the copy constructor at all. I did some more testing with this and found that the original version behaves properly.

PR #158, opened 2016-08-14
Fixed MESOS-5925 Building Mesos in Docker 1.12 on OS X

haosdent:
- Replaced gmock-1.7.0.tar.gz with a correct permissions version.
- Replaced gperftools-2.5.tar.gz with a correct permissions version.

Because I found upload binary changes does not work in https://reviews.apache.org/r/51079/ and  https://reviews.apache.org/r/51080/,  I open this pull request.


haosdent:
cc @tillt @karya0 @radekg

radekg:
@haosdent yep, this solves the problem, I guess I can fix my pr?

haosdent:
@radekg Yes, I think to change the incorrect permissions in the tar packages is better than add `--no-same-owener` as a workaround.

radekg:
👍  from me to merge this

PR #159, opened 2016-08-19
MESOS-6056 add NOOP Container Logger for mesos.

IvanJobs:
mesos has two Container Loggers in its source files. 
One is build into mesos-agent: sandbox Container Logger, it just redirects stderr/stdout to sandbox, causing fill disk usage problem.
The other is LogrotateContainerLogger module lib, it's good, we can make sure stdout/stderr in sandbox be in a constant size.
But there is a common need: don't write stdout/stderr into sandbox, pity, we don't have any flags for turning it off. 

This is a come around for this: developing a new module lib for ContainerLogger for doing nothing(redirect stdout/stderr to /dev/null)

yep, that's it. We need a NOOP ContainerLogger,.

### DIFF
--- a/docs/contributors.yaml
+++ b/docs/contributors.yaml
@@ -531,3 +531,9 @@
     - w.rouesnel@gmail.com
   jira_user: wrouesnel
   reviewboard_user: wrouesnel
+
+ - name: Ivan Jobs
+  emails:
+    - ivan1377@163.com
+  jira_user: IvanJobs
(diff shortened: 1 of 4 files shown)


IvanJobs:
Well, actually after communication with Joseph Wu, I think this NOOP Container Logger is not so common and should not be accept by mesos community. So just forget about it. But if you have special use case and want to use this, I'm happy about that

PR #162, opened 2016-08-26
Frameworks UI shows metrics for used resources plus offers

bernadinm:
https://issues.apache.org/jira/browse/MESOS-6098

When a framework is receiving many offers and it is denying them, the frameworks UI will show the metrics fluctuating for mem, cpu, gpu, and disk.
From a mesos perspective, those offers are given to the framework until the framework declines them, so depending on the time the mesos UI gets updated, its has combined all the used resources and offers (that have not been accepted) to the framework and is reflected on the framework UI. If a framework does not implement suppressiveOffers(), it will continue to deny offers from mesos, which leads to the sporadic changes of metrics on the framework UI.
From the operator's perspective, the user would expect to see used resources consumed by the framework. Any offered resources can be viewed instead by Mesos's Offers tab.

cc: @kaysoky

### DIFF
--- a/src/webui/master/static/frameworks.html
+++ b/src/webui/master/static/frameworks.html
@@ -16,10 +16,10 @@
       <th data-key="role">Role</th>
       <th data-key="principal">Principal</th>
       <th data-key="tasks.length">Active Tasks</th>
-      <th data-key="resources.cpus">CPUs</th>
-      <th data-key="resources.gpus">GPUs</th>
-      <th data-key="resources.mem">Mem</th>
-      <th data-key="resources.disk">Disk</th>
+      <th data-key="used_resources.cpus">CPUs</th>
+      <th data-key="used_resources.gpus">GPUs</th>
+      <th data-key="used_resources.mem">Mem</th>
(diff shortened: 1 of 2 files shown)


kaysoky:
LGTM.

Checked the UI by running a couple agents:
```
bin/mesos-master.sh --work_dir=/tmp/master --ip=127.0.0.1
bin/mesos-agent.sh --master=localhost:5050 --work_dir=/tmp/agent1
bin/mesos-agent.sh --master=localhost:5050 --work_dir=/tmp/agent2 --port=5052
bin/mesos-agent.sh --master=localhost:5050 --work_dir=/tmp/agent3 --port=5053
```
Commenting out this line: 
https://github.com/apache/mesos/blob/fb23845214088bf48dd0e9e0324cafa3643d574a/src/cli/execute.cpp#L453

Then building and running:
```
src/mesos-execute --master=localhost:5050 --command="sleep 1000" --name="Hoarding task"
```

As expected, the UI showed 1 CPU of usage, consistently.  Checking the "Offers" tab showed one offer for each agent (hoarded).

PR #164, opened 2016-09-02
Getting started Doc updated.

khushbuparakh:
openjdk-7-jdk is obsolete with updated version of ubuntu OS 16.04 so commands and dependencies  are updated so that it can be suitable for both versions xenial and trusty

### DIFF
--- a/3rdparty/stout/include/stout/windows.hpp
+++ b/3rdparty/stout/include/stout/windows.hpp
@@ -50,11 +50,6 @@
 #error "Mesos doesn't currently support the `_UNICODE` Windows header constant"
 #endif // _UNICODE
 
-// Similarly, the Windows API uses `PWSTR` to mean `wchar_t*`,
-// but rather than assuming this is true, it should be asserted.
-static_assert(std::is_same<PWSTR, wchar_t*>::value,
-              "Expected `PWSTR` to be of type `wchar_t*`.");
-
 // An RAII `HANDLE`.
 class SharedHandle : public std::shared_ptr<void>
 {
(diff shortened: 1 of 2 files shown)


support/docker_build.sh
```diff
@@ -63,7 +63,7 @@ case $OS in
     [ "$(uname -m)" = "x86_64" ] && CLANG_PKG=clang-3.5 || CLANG_PKG=
     append_dockerfile "RUN apt-get update"
     append_dockerfile "RUN apt-get -y install build-essential $CLANG_PKG git maven autoconf libtool cmake"
-    append_dockerfile "RUN apt-get -y install openjdk-7-jdk python-dev libcurl4-nss-dev libsasl2-dev libapr1-dev libsvn-dev libevent-dev libev-dev"
+    append_dockerfile "RUN apt-get -y install openjdk-8-jdk python-dev libcurl4-nss-dev libsasl2-dev libapr1-dev libsvn-dev libevent-dev libev-dev"
```
haosdent: Ubuntu 14.04 din't contain openjdk-8-jdk in their repo.

docs/getting-started.md
```diff
@@ -30,7 +30,7 @@ Mesos agent also run on Windows. To build Mesos agent from source, Visual Studio
 
 Make sure your hostname is resolvable via DNS or via `/etc/hosts` to allow full support of Docker's host-networking capabilities, needed for some of the Mesos tests. When in doubt, please validate that `/etc/hosts` contains your hostname.
 
-### Ubuntu 14.04
+### Ubuntu 16.04
```
haosdent: We still need to support 14.04, may you add a new section only for 16.04?

janisz:
Looks like this was resolved by #173

PR #165, opened 2016-09-02
mesoscon eu - hackatron exercise - CI using travis

dcaba:
Hi,

  reusing some of the work my colleague @danigiri did (mesos-build-helper) , i've adapted (and incorporated into the project) some scripts that, inside a container:
- resolves the compilation dependencies
- compiles
- and execute the tests
  
  So TravisCI can execute this on branches and when triggering PRs against the project (currently is active, but does nothing out of adding red flags)...

Pending to do:
- actual error checking (probably we want to break the build if make test fails)
- remove some references to mesos_version (current scripts pick the latest checked out version)
  
  That's it for now... let me know if it makes sense to complete the work (you probably are covering CI in a different way), but having this integrated in github may be good for contributors. Example of job that travis executed:
  https://travis-ci.org/dcaba/mesos/builds/157089050


Dockerfile-ci
```diff
@@ -0,0 +1,23 @@
+#  Copyright 2015 Daniel Giribet <dani - calidos.cat>
```
janisz: Can we use same Dockerfile as we are using on Jenkins? Take a look at [support/docker_build.sh](https://github.com/apache/mesos/blob/42e8d6fb8b894c08b4b0c7cc40036fe630489bba/support/docker_build.sh)

dcaba: @janisz , yes! Not aware of its existence, and this duplication makes no sense... probably we should also add a link to http://mesos.apache.org/documentation/latest/submitting-a-patch/ in the project README, so all these supporting services are known for new contributors to apache projects. I will add a couple of commits.

dcaba:
The build times out after 50 minutes (maximum job length the free version of travis allows)... I'm afraid we cannot afford executing all the tests in Travis (last one that initiates is tests/health_check_tests.cpp), so I can make configurable in support/docker_build.sh the task in the Makefile we want to execute (being the distcheck the default one, that Jenkins will continue using), or just dismiss Travis usage for this project... @janisz , any ideas/recommendations about how to proceed?

kaysoky:
You can see [on the ASF Mesos CI](https://builds.apache.org/job/Mesos/buildTimeTrend) that successful builds typically takes 60-90 minutes.  However, this is for `make distcheck`.  

A `make check` should take ~40 minutes on a c3.2xlarge AWS instance (8 cores, 15 GB RAM, SSDs).  This is probably beyond the free-tier of Travis.

If you're interested, here are two commits that show the history of Travis CI usage in Mesos.  It was only enabled briefly between 0.22 and 0.23 :smile: :
* https://reviews.apache.org/r/28683/
* https://reviews.apache.org/r/34225/

haosdent:
Some experience to make it faster as well. https://reviews.apache.org/r/39257/diff/1#index_header (`- CXXFLAGS=-gsplit-dwarf`)
But I think we give up travis ci long time ago? Refer to https://issues.apache.org/jira/browse/MESOS-5655?focusedCommentId=15339918&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-15339918

dcaba:
Hi @kaysoky , @haosdent !

  thanks so much for your comments. Yep, I see this is not new for the project, and I was missing a lot of context when I initiated this PR. So now we have two options, from my point of view:
- Dismiss this PR. This integration brings small or no value to the project (at a cost of a .travis.yml and keeping the integration active), as the free version of travis does not meet all the CI requirements you currently have covered with Jenkins.
- Accept the PR. This will fix the "red flag" we can see in all the PRs, and this job will potentially provide quicker feedback as we are only compiling in the end (while Jenkins will bring the results of all the test coverage later)

   Please, make the decision bearing in mind what's more convenient for the project... No rancor at all if you close without merge ;) (it was an exercise requiring small efforts). But I'd appreciate the related Jira issue is closed (MESOS-5655) an you actually disable Travis to prevent these red crosses to add noise to all PRs.

    Best,

dcaba:
BTW,

   I've just remembered a benefit of keeping this active (that will not affect too much that decision); Travis will run the compilation / ci task not only when you raise a PR against the mesos repo, but also in any branch or fork if the account has the integration active (not sure how Jenkins is configured in your case, but probably it does not have visibility of the forks potential contributors may be initating).

    Best!

janisz:
I agree with @dcaba if Travis is enabled then it should at least run style check (whole compile is good too).
> But I think we give up travis ci long time ago?
Why it's still enabled?

janisz:
+1 for merge. Can someone shepherd it or disable Travis and wire ASF Jenkins instead?

dcaba:
Hi,

  a bit surprised travis integration is still enabled (so all PRs have red crosses)... let me know if you want me to resolve the conflicts to integrate this (in addition to a basic travis job, there's also some other small amends I think are interesting). If not, I will just close.

haosdent:
Hi, @dcaba Since Mesos have used Apache CI to perform testings instead of travis ci, could refer to https://issues.apache.org/jira/browse/MESOS-5655 for the details. And thx @jfarrell, he just disabled the travis ci for Mesos. cc @vinodkone

dcaba:
Great. Thanks for the update; this parallel CI job in travis (this PR was bringing) was not adding too much benefits (just the nice github <-> travis integration, that can show green ticks or red crosses in PRs submitted in this repo?), and it was only worth to fix the red crosses while the integration was still enabled. Closing PR then! Best!

haosdent:
yah, actually apache jenkins have integrated with github. Could provide build status via this badge icons.

Build status: 
cmake,gcc,--verbose --enable-libevent --enable-ssl,GLOG_v=1 MESOS_VERBOSE=1,centos:7 &&(!ubuntu-us1)&&(!ubuntu-eu2)/badge/icon)

PR #168, opened 2016-09-19
Update contributors.yaml for gradywang.

gradywang:
Update contributors.yaml for gradywang.

### DIFF
--- a/docs/contributors.yaml
+++ b/docs/contributors.yaml
@@ -402,13 +402,14 @@
   jira_user: yanyanhu
   reviewboard_user: yanyanhu
 
-- name: Yong Qiao Wang
+- name: Yongqiao Wang
   affiliations:
-    - {organization: IBM}
+    - {organization: Huawei, start_date: 2016-09-15}
+    - {organization: IBM, end_date: 2016-09-14}
   emails:
-    - yqwyq@cn.ibm.com
+    - grady.wang@foxmail.com
   jira_user: gradywang
-  reviewboard_user: JamesYongQiaoWang
+  reviewboard_user: gradywang
 
 - name: Yong Tang
   emails:


docs/contributors.yaml
```diff
@@ -402,13 +402,13 @@
   jira_user: yanyanhu
   reviewboard_user: yanyanhu
 
-- name: Yong Qiao Wang
+- name: Yongqiao Wang
   affiliations:
-    - {organization: IBM}
```
vinodkone: Instead of removing your old affiliation (IBM) completely, can you add an "end_date" to it? That way your contributions while you were employed at IBM, will be attributed correctly. See my affiliation above for an example.

gradywang: Thank @vinodkone, I have fixed by your comments.

gradywang:
@vinodkone, cloud you help to review this again? Thanks!

docs/contributors.yaml
```diff
@@ -402,13 +402,14 @@
   jira_user: yanyanhu
   reviewboard_user: yanyanhu
 
-- name: Yong Qiao Wang
```
vinodkone: If your name and reviewboard username have also changed along with your affiliation, then it's probably worth to have two different entries in the YAML file.

```
- name: Yong Qiao Wang
  affiliations:
    - {organization: IBM}
  emails:
    - yqwyq@cn.ibm.com
  jira_user: gradywang
  reviewboard_user: JamesYongQiaoWang

- name: Yongqiao Wang
  affiliations:
    - {organization: Huawei}
  emails:
    - grady.wang@foxmail.com
  jira_user: gradywang
  reviewboard_user: gradywang
```

gradywang: @vinodkone I also used `gradywang` as my reviewboard username  rather than `JamesYongQiaoWang` when I in IBM in main contributions. For my name, it does not changed, and just change to the recommended format. So can I only change the old entry?

vinodkone: oh ok. then a single entry sounds fine.

PR #169, opened 2016-09-20
Add smarter master redirects.

drcrallen:
* Fix MESOS-6210
* Any path after `/redirect` is now appended to the response
* Does not make any guarantees about non-path components of
the original request
* Updated docs to clarify this behavior.


drcrallen:
This was supposed to be agains the metamx branch, not apache. Will open official contribution once internal stuff is finished.

PR #184, opened 2016-11-21
In CentOS 7, the libevent-devel dependency is needed to be installed for mesos SSL build.

chjhaijiang:
In  the link (https://github.com/apache/mesos/blob/master/docs/getting-started.md), it should be a proper instruction once user want to build a SSL mesos. Right now, it is missed.


xiaods:
OMG, it is need rebase.

PR #192, opened 2016-12-07
Fix python dependencies to avoid installed newer version

chengweiv5:
Without PACKAGE_VERSION specified, it may install newly python libs from  pypi
if there is one.

Signed-off-by: Chengwei Yang <yangchengwei@qiyi.com>

### DIFF
--- a/src/python/setup.py.in
+++ b/src/python/setup.py.in
@@ -26,8 +26,8 @@ config = {
     'packages': [ 'mesos' ],
     'package_dir': { '': 'src' },
     'install_requires': [
-        'mesos.interface',
-        'mesos.native'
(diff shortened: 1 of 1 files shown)


chengweiv5:
@haosdent Please help to review

haosdent:
Refer to 
* https://github.com/apache/mesos/blob/1.1.0/src/python/native/setup.py.in#L29-L30
* https://github.com/apache/mesos/blob/1.1.0/src/python/scheduler/setup.py.in#L31
* https://github.com/apache/mesos/blob/1.1.0/src/python/executor/setup.py.in#L31
This patch looks good to me. @tillt May you help review this patch again? Thank you in advance.

PR #198, opened 2017-01-12
Fix inconsistency in the port_mapping.cpp documentation

thomas-maurice:
The change is very small and self explanatory.

This inversion between tx & rx can lead to confusions


jieyu:
Can you split this into two PR? One for adding yourself into contributor list, the other one for the actual fix?

thomas-maurice:
Done !
The second PR is #199

PR #205, opened 2017-03-04
Clarify existence of FrameworkID in SUBSCRIBE calls.

aaronjwood:
I wanted to clarify how `FrameworkID` should or shouldn't be set in `SUBSCRIBE` calls. Spoke with Anand about this earlier today when I was running into issues with resubscribing.


hatred:
Thanks for the PR. I couldn't find a way to assign myself as a reviewer here. I would review this and help get it committed.

aaronjwood:
@hatred No problem, thanks for the help!

vinodkone:
@hatred still planning to review this?

aaronjwood:
@neilconway you had mentioned a while back that this could be clarified better. Can you elaborate on what you were thinking so we can get this merged in?

greggomann:
@vinodkone could you take a look?

PR #224, opened 2017-07-25
Bundled gRPC v1.4.2 into 3rdparty libraries.

chhsia0:
The bundled package is generated with the following command:

```sh
git clone -b v1.4.2 https://github.com/grpc/grpc.git grpc-1.4.2
(cd grpc-1.4.2 && git submodule update --init third_party/cares)
tar zcvf grpc-1.4.2.tar.gz --exclude .git grpc-1.4.2
```

We download gRPC from GitHub instead of using the official tarball
because we need to use `git submodule` to download dependent third-party
packages for gRPC, and the necessary git metadata does not exist in the
official tarball.

gRPC also depends on openssl and protobuf 3. Since we have already
bundled the latter, the only system dependency is openssl.


janisz:
Link: [MESOS-7808](https://issues.apache.org/jira/browse/MESOS-7808)

chhsia0:
Closing it as it is already committed from reviewboard.

PR #233, opened 2017-09-13
Display task health in the Web UI.

janisz:
Fixes: [MESOS-7961](https://issues.apache.org/jira/browse/MESOS-7961)

screencapture-localhost-5050-1505330708402

### DIFF
--- a/src/webui/master/static/home.html
+++ b/src/webui/master/static/home.html
@@ -162,6 +162,7 @@
           <th data-key="name">Task Name</th>
           <th data-key="role">Role</th>
           <th data-key="state">State</th>
+          <th data-key="healthy">Health</th>
           <th data-key="start_time" data-sort>Started</th>
           <th data-key="host">Host</th>
           <th></th>
@@ -169,7 +170,7 @@
       </thead>
       <tbody>
         <tr data-ng-if="active_tasks.length === 0">
-          <td colspan="7">No active tasks.</td>
+          <td colspan="8">No active tasks.</td>
         </tr>
         <tr ng-repeat="task in $data">
           <td>
@@ -185,6 +186,7 @@
           <td>{{task.name}}</td>
           <td>{{task.role}}</td>
           <td>{{task.state | truncateMesosState}}</td>
+          <td class="task-{{task.healthy | taskHealth}}">{{task.healthy | taskHealth}}</td>
           <td>
             <m-timestamp value="{{task.start_time}}"></m-timestamp>
           </td>
--- a/src/webui/master/static/js/app.js
+++ b/src/webui/master/static/js/app.js
@@ -88,6 +88,14 @@
         return state.substring(5);
       };
     })
+    .filter('taskHealth', function() {
+      return function(health) {
+        if (health != null) {
+           return health ? "healthy" : "unhealthy";
+        }
+        return "-";
+      }
+    })
     .filter('isoDate', function($filter) {
       return function(date) {
         var i = parseInt(date, 10);
--- a/src/webui/master/static/js/controllers.js
+++ b/src/webui/master/static/js/controllers.js
@@ -156,6 +156,7 @@
           if (isStateTerminal(task.state)) {
               task.finish_time = lastStatus.timestamp * 1000;
           }
+          task.healthy = lastStatus.healthy;
       }
     };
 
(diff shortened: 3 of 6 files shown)


lilysamimi:
Would be interested in talking with the folks that created this UI component. I'm giving out $10 Amazon gift cards to individuals who participate in a user experience research study with me pertaining to the Mesos UI. Please DM me. I'm at MesosCon today and tomorrow and around the Mesosphere booth.

bmahler:
Thanks janisz, this looks pretty good. However, this needs to be added to several places:

- The global tasks table (you covered this)
- The framework-specific task table (when clicking into a framework)
- The executor-specific task table within an agent (when clicking into an agent, then a framework, then an executor)

I'm also a bit curious about whether it's a good idea to re-use the health string as a class, since it seems easy to break vs only specifying the display string in the html page.

Will you be able to continue working on this?

janisz:
@bmahler Thanks for the review. Most of the issues are minor and easy to fix. I do it shortly.

Vish0007:
@janisz  the UI looks good and health check does add value.
would you describe how the "Task Name" and "Health" are different, i can see health status in both places.

janisz:
@Vish0007 To create this screenshot I created 3 applications called `unhealthy`, `healthy` and `unknown` so it's just a coincidence that the name has the same value as tasks health. When I fix all issues I'll prepare gif to present healthy is available in all places and I'll use more generic names in order to avoid confusion.

Vish0007:
@janisz sounds great.

src/webui/master/static/home.html
```diff
@@ -162,14 +162,15 @@
           <th data-key="name">Task Name</th>
```
Vishant0031: is this task id (MESOS_TASK_ID?), should this be "Task Id"?

janisz: Task ID is one line above

janisz:
* [x] The global tasks table
* [x] The framework-specific task table (when clicking into a framework)
* [x] The executor-specific task table within an agent (when clicking into an agent, then a framework, then an executor)

> I'm also a bit curious about whether it's a good idea to re-use the health string as a class, since it seems easy to break vs only specifying the display string in the html page.

I removed CSS. I'm not sure if red/green looks good in Mesos blue palette. @lilysamimi Any idea how to present health state?


health



<details>

I used Marathon to deploy 3 applications with 3 types of health state.

```
 curl -X PUT -d@health_groups.json localhost:8080/v2/groups/
```

```json
{
   "apps":[
      {
         "id":"/apache",
         "cmd":"sleep 1000",
         "cpus":1,
         "mem":128,
         "disk":0,
         "instances":1,
         "healthChecks":[
            {
               "gracePeriodSeconds":1,
               "intervalSeconds":1,
               "maxConsecutiveFailures":300,
               "command":{
                  "value":"true"
               },
               "protocol":"COMMAND"
            }
         ]
      },
      {
         "id":"/tomcat",
         "cmd":"sleep 1000",
         "cpus":1,
         "mem":128,
         "disk":0,
         "instances":1,
         "healthChecks":[
            {
               "gracePeriodSeconds":1,
               "intervalSeconds":1,
               "maxConsecutiveFailures":300,
               "command":{
                  "value":"false"
               },
               "protocol":"COMMAND"
            }
         ]
      },
      {
         "id":"/mysql",
         "cmd":"sleep 1000",
         "cpus":1,
         "mem":128,
         "disk":0,
         "instances":1
      }
   ]
}
```

</details>

bmahler:
@janisz I noticed you updated the patch, is it ready for another review? Would be great to keep the colors as you had before. I was just asking about the approach used :)

Also can you add a gif or video of your screen to show the testing of it?

janisz:
@bmahler I added some colours. 

health

janisz:
@kwiatkk1 Thanks for the review. I applied your suggestions.
@bmahler Changed `Unknown` to `-`


out-2

src/webui/master/static/js/controllers.js
```diff
@@ -838,6 +839,15 @@
             });
           }
 
+          function setHealth(tasks) {
+            _.each(tasks, function(task) {
+              var lastStatus = _.last(task.statuses);
+              if (lastStatus) {
+                  task.healthy = lastStatus.healthy;
```
bmahler: 2 space indents

src/webui/master/static/js/app.js
```diff
@@ -88,6 +88,14 @@
         return state.substring(5);
       };
     })
+    .filter('taskHealth', function() {
+      return function(health) {
+        if (health != null) {
+           return health ? "healthy" : "unhealthy";
```
bmahler: 2 space indent, would also be good to warn the reader here that this value is relied on for the CSS

PR #234, opened 2017-09-13
Display task state counters in the framework page.

janisz:
Fixes: [MESOS-7962](https://issues.apache.org/jira/browse/MESOS-7962)

screencapture-localhost-5050-1505334897521

### DIFF
--- a/src/webui/master/static/js/controllers.js
+++ b/src/webui/master/static/js/controllers.js
@@ -230,6 +230,34 @@
       _.each(framework.unreachable_tasks, setTaskMetadata);
       _.each(framework.completed_tasks, setTaskMetadata);
 
+      framework.running_tasks = 0;
+      framework.staging_tasks = 0;
+      framework.starting_tasks = 0;
+      framework.killing_tasks = 0;
+
+      _.each(framework.tasks, function(task) {
+        switch (task.state) {
+            case "TASK_RUNNING": framework.running_tasks++; break;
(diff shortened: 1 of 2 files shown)


lilysamimi:
Would be interested in talking with the folks that created this UI component. I'm giving out $10 Amazon gift cards to individuals who participate in a user experience research study with me pertaining to the Mesos UI. Please DM me. I'm at MesosCon today and tomorrow and around the Mesosphere booth.

src/webui/master/static/js/controllers.js
```diff
@@ -230,6 +230,34 @@
...
       _.each(framework.completed_tasks, setTaskMetadata);
 
+      framework.running_tasks = 0;
+      framework.staging_tasks = 0;
+      framework.starting_tasks = 0;
+      framework.killing_tasks = 0;
+
+      _.each(framework.tasks, function(task) {
+        switch (task.state) {
+            case "TASK_RUNNING": framework.running_tasks++; break;
```
bmahler: Let's put running after starting.

src/webui/master/static/js/controllers.js
```diff
@@ -230,6 +230,34 @@
       _.each(framework.unreachable_tasks, setTaskMetadata);
       _.each(framework.completed_tasks, setTaskMetadata);
 
+      framework.running_tasks = 0;
```
bmahler: A TODO here for adding per-framework metrics so that we don't have to loop over the tasks would be great.

PR #237, opened 2017-10-03
Documentation: Fix event syntax by wrapping a task definition in curly braces.

adamtheturtle:
cc @greggomann

### DIFF
--- a/docs/executor-http-api.md
+++ b/docs/executor-http-api.md
@@ -279,19 +279,21 @@ LAUNCH_GROUP Event (JSON)
   "launch_group": {
     "task_group" : {
       "tasks" : [
-        "task": {
-          "name": "dummy-task",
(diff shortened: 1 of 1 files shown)


docs/executor-http-api.md
```diff
@@ -279,19 +279,21 @@ LAUNCH_GROUP Event (JSON)
...
-          "agent_id": {
-            "value": "f1c9cdc5-195e-41a7-a0d7-adaa9af07f81"
-          },
-          "command": {
-            "value": "sleep",
-            "arguments": [
-              "100"
-            ]
+        {
+          "task": {
```
rukletsov: Shouldn't we remove `"task": ` instead?

adamtheturtle:
Good spot @rukletsov . I have updated this as appropriate.

PR #240, opened 2017-10-06
Corrected quota to define it as minimal and maximal

aggress:
Based on a discussion with @bmahler, the current documented definition of a quota suggests it guarantees only the minimal guarantee, and that a framework can burst above that minimal.  This is apparently incorrect today, so I've updated the definition accordingly.

### DIFF
--- a/docs/quota.md
+++ b/docs/quota.md
@@ -49,7 +49,8 @@ In computer science, a "quota" usually refers to one of the following:
 * A pair of both.
 
 In Mesos, a quota is a **guaranteed** resource allocation that a role may rely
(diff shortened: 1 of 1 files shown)


rukletsov:
Is this comment insufficient? https://github.com/apache/mesos/blame/c87db8c8c8cef3d95e8ddaf7c75353732919edf7/docs/quota.md#L321
Maybe pull it higher then?

aggress:
@rukletsov yes, it's insufficient. The initial description is incorrect and should be corrected. That other statement is way, way down beyond the cut where users are unlikely to venture.

PR #243, opened 2017-10-25
Update home.md

packtpartner:
Added new book listing - Apache Mesos Cookbook


jpeach:
Markup-wise this looked fine to me, but the grid didn't lay out correctly (this new addition appends to the far right on a new row). Maybe someone familiar with bootstrap can figure it out?

janisz:
The problem is because images have a different height (one have different ratio). We can fix this by scaling them. Just adding 
```css
max-height: 200px;
```
 to book thumbnail should be enough.

jpeach:
Thanks @janisz!

PR #244, opened 2017-10-28
Add SliceLife in powered-by-mesos.md

bobeckert:
Proudly using Mesos!


bmahler:
Thanks @bobeckert! This is now committed.

PR #245, opened 2017-11-01
Add Recombee to the powered by list.

deric:
We've been using Mesos for while, it's time to add Recombee here.


bmahler:
Thanks @deric!

PR #255, opened 2018-01-12
Add curl examples to operator-http-api

nhandler:
When reading this document, many people will want to test out the
requests against their own environment. Providing curl examples will
help assist with this task.

### DIFF
--- a/docs/operator-http-api.md
+++ b/docs/operator-http-api.md
@@ -28,23 +28,19 @@ Below are the example calls to master that result in synchronous responses from 
 This call retrieves the health status of master.
 
 ```
-GET_HEALTH HTTP Request (JSON):
-
-POST /api/v1  HTTP/1.1
-
-Host: masterhost:5050
-Content-Type: application/json
-Accept: application/json
-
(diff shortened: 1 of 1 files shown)


vinodkone:
@nhandler  Thanks for doing this! Awesome. Do I take it that you tested all these commands? Also, is the EOF required in all the curl commands?

vinodkone (approved):
LGTM!

greggomann:
@nhandler similar to Vinod's comment: I think we could probably get away without the heredoc notation, to get rid of the EOFs?

i.e.:
```
curl -i \
  -X POST http://<masterhost>:5050/api/v1 \
  -H "Content-Type: application/json" \
  -H "Accept: application/json" \
  -d '
{
  "type": "GET_HEALTH"
}'
```
This works when I test it in bash anyway :)

vinodkone:
@nhandler If you can update the PR, I'll commit it ASAP.

greggomann:
@nhandler we were reviewing PRs today and came across this one - do you have a cycle to update real quick?

PR #258, opened 2018-01-18
Update api-client-libraries.md

miroswan:
* Adding Go client for the Mesos Operator API


vinodkone:
Thanks for the submission! Are you planning to use this in production?

miroswan:
Yep. I operate a Mesos cluster and plan on using this client to make updates to resource reservations, but I believe there will be some other use cases down the line.

PR #259, opened 2018-01-22
Update contributors.yaml

AntonDan:
Entry for myself in the contributors file


docs/contributors.yaml
```diff
@@ -777,3 +777,9 @@
     - xingzhou@cn.ibm.com
   jira_user: dongdong
   reviewboard_user: dongdong
+  
+- name: Antonis Danezis
```
jieyu: Can you sort the entry based on first name? Thanks!

AntonDan: Ah my bad, didn't realize they were sorted. Hope it's fixed now

PR #262, opened 2018-02-05
Updating reservation.md

miroswan:
When using the Operator API, I came to this document to get a better understand of resource reservations. It wasn't clear to me that the reservation or reservations keys were a requirement given that the principal configuration is optional. What I do know is that the role can be set in the resource object or the reservation object. I think we might as well instruct users to set the role in the reservation object so it is more clear that it is required to successfully POST the request. After a quick google search, I noted that another user ran into the same issue and posted about it in IRC.

### DIFF
--- a/docs/reservation.md
+++ b/docs/reservation.md
@@ -623,18 +623,18 @@ We send an HTTP POST request to the master's
               "name": "cpus",
               "type": "SCALAR",
               "scalar": { "value": 8 },
-              "role": "ads",
               "reservation": {
-                "principal": <operator_principal>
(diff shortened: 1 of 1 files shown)


vinodkone:
@mpark @bmahler Can you guys review?

bmahler:
@mpark I didn't realize we also supported setting role within `reservation`, should this be adding to `reservations`? Or do we officially support `reservation` with type `DYNAMIC` & `role`?

mpark:
7cee22c

PR #263, opened 2018-02-09
Allow nested containers in pods to have separate namespaces(Ref: MESOS-8534).

sagar8192:
This change allows nested containers to have separate network and mount namespaces. It also retains the existing functionality, where a nested container can attach to parent's network and mount namespace.

I have not fixed/added tests for this. First, I want to make sure that this is the right approach. If this change looks good to the reviewers, I will fix/add unit tests.

After this change is shipped, the docs also need to be updated.

### DIFF
--- a/src/slave/containerizer/mesos/isolators/network/cni/cni.cpp
+++ b/src/slave/containerizer/mesos/isolators/network/cni/cni.cpp
@@ -570,10 +570,17 @@ Future<Option<ContainerLaunchInfo>> NetworkCniIsolatorProcess::prepare(
     return Failure("Container has already been prepared");
   }
 
+  bool needsSeparateNs = false;
+  if ((containerConfig.has_container_info() &&
+        containerConfig.container_info().network_infos().size() > 0) ||
+            !containerId.has_parent()) {
+    needsSeparateNs = true;
+  }
+
   hashmap<string, ContainerNetwork> containerNetworks;
   Option<string> hostname;
 
-  if (!containerId.has_parent()) {
+  if (needsSeparateNs) {
     const ExecutorInfo& executorInfo = containerConfig.executor_info();
     if (!executorInfo.has_container()) {
       return None();
@@ -618,13 +625,6 @@ Future<Option<ContainerLaunchInfo>> NetworkCniIsolatorProcess::prepare(
     // `ContainerConfig.container` is set, it implies that the nested
     // container needs to have a separate network namespace, else the
     // nested container shares its network namespace with the parent.
-    if (containerConfig.has_container_info() &&
-        containerConfig.container_info().network_infos().size() > 0) {
-      return Failure(
-          "Currently, we don't support different network namespaces for "
-          "parent and nested containers.");
-    }
-
     ContainerID rootContainerId = protobuf::getRootContainerId(containerId);
 
     // NOTE: The `network/cni` isolator checkpoints only the following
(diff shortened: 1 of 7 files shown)


src/slave/containerizer/mesos/isolators/network/cni/cni.cpp
```diff
@@ -820,18 +821,16 @@ Future<Nothing> NetworkCniIsolatorProcess::isolate(
   CHECK_SOME(rootDir);
   CHECK_SOME(pluginDir);
 
-  if (containerId.has_parent()) {
+  if (!infos[containerId]->needsSeparateNs) {
```
jieyu: I'd make this more explicit.
```
// NOTE: DEBUG container should not have Info struct. Thus if the control
// reaches here, the container is not a DEBUG container.
if (isNestedContainer && joinParentNetwork)
```

src/slave/containerizer/mesos/isolators/network/cni/cni.cpp
```diff
@@ -751,10 +751,11 @@ Future<Option<ContainerLaunchInfo>> NetworkCniIsolatorProcess::prepare(
         launchInfo.add_clone_namespaces(CLONE_NEWNET);
         launchInfo.add_clone_namespaces(CLONE_NEWNS);
         launchInfo.add_clone_namespaces(CLONE_NEWUTS);
+        infos[containerId]->needsSeparateNs = needsSeparateNs;
```
jieyu: I'd just store `joinParentNetwork` boolean in the `Info` struct.

Looks like we need some way to checkpoint this information. Otherwise, after recovery, how does the isolator tell if it needs to call CNI detach for nested container? Currently, since nested container always share with its parent, no need to do any cleanup. But it's no longer true with this change.

But I think we can tell if we discover a nested container having checkpointed data under `/var/run/mesos/isolators/network/cni/`. See `src/slave/containerizer/mesos/isolators/network/cni/paths.hpp` for the checkpointing layout. You'll notice this when you properly implement recover() method.

jdef:
What's the high level use case that's driving this change request? One of the major goals of task-groups (pods) is to allow containers to share networking and storage. What point is there in launching a nested container that DOES NOT share these things with the other containers in the pod?

Gilbert88:
A quick note that we could have a followup patch to add documents here: http://mesos.apache.org/documentation/latest/containerizer-internals/#linux-namespaces

qianzhangxa:
I'd like to echo @jdef's comment, we need a clear use case for ip per nested container. Our current status is, if framework launches multiple task groups (pods) via a single default executor, all the nested containers of all these task groups will share the executor's network namespace. This is actually different from Kubernetes pod where each pod will have its own network namespace and all the container in a pod will share the same network namespace so that they can communicated with 127.0.0.1/localhost. IMHO, we should consider to do something similar with Kubernetes, i.e., each task group will have its own network namespace rather than each nested container has its own network namespace unless we have a use case for it.

sagar8192:
@jdef, @qianzhangxa: I have added some more information about the use case to the [ticket](https://issues.apache.org/jira/browse/MESOS-8534). Please check it out. We are planning to discuss this in today's containerization sync.

sagar8192:
Posted a new review here: https://reviews.apache.org/r/65987/

sagar8192:
Closed in favor of https://reviews.apache.org/r/65987/

PR #266, opened 2018-02-22
Tasks docs

kohend:
Indicate that filtering is a new feature.

### DIFF
--- a/docs/endpoints/master/tasks.json.md
+++ b/docs/endpoints/master/tasks.json.md
@@ -24,6 +24,8 @@ Lists known tasks.
 The information shown might be filtered based on the user
 accessing the endpoint.
 
+Filtering via the query parameters is available starting Mesos version 1.4.0.
+
 Query parameters:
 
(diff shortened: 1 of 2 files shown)


bmahler:
We've removed these docs entirely from the repo since they were very stale and they now get generated as part of the website publishing, you can find them here:

http://mesos.apache.org/documentation/latest/endpoints/

The way we will address what you're adding here is to provide versioned documentation, that work is ongoing here: https://reviews.apache.org/r/52064/ cc @tanderegg

bmahler:
@kohend can you close this PR based on my comment above?

kohend:
Is it relevant to add it to the documentation in the source instead, so that the difference will have better visibility (and people won't have to go through versions to see when it's added)?

bmahler:
@kohend Can you clarify? I'm not sure what you mean.

greggomann:
Closing due to inactivity - reopen if still relevant, thanks!

PR #278, opened 2018-03-27
Clarified usage of docker_remove_delay flag.

joerg84:


joerg84:
@Gilbert88 As discussed

PR #279, opened 2018-03-30
WIP: Remove unknown unreachable tasks when agent re-registers.

m9a:
A RunTask messsage could get dropped for an agent while it's
disconnected from the master and when such an agent goes unreachable
then this dropped task message gets added to the unreachable tasks.
When the agent re-registers, the master sends status updates for the
tasks that the agent reported when re-registering and these tasks are
also removed from the unreachableTasks on the framework but since the
agent doesn't know about the dropped task so it doesn't get removed
from the unreachableTasks leading to a check failure when
this inconsistency is detected during framework removal.

JIRA for this PR: https://issues.apache.org/jira/browse/MESOS-8750

### DIFF
--- a/src/master/master.cpp
+++ b/src/master/master.cpp
@@ -1926,6 +1926,7 @@ void Master::_doRegistryGc(
     }
 
     slaves.unreachable.erase(slave);
+    slaves.unreachableAgentsToTasks.erase(slave);
     numRemovedUnreachable++;
   }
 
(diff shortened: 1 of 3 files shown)


Gilbert88:
Hi @m9a , thanks for the PR. While the issue is still in WIP stage. Would you mind to create a [JIRA-where people track on issues](https://issues.apache.org/jira/projects/MESOS) and post a patch at [reviewboard - required by Apache](http://reviews.apache.org/)? Usually, we only accept document PR or small change via Github. It was a practice from Apache. Sorry for the inconvenience.

xujyan:
@Gilbert88: @m9a was having difficulty in with rbt so I told her to use a PR. Let's see how it goes first. @m9a have you created a JIRA and are you able to set the reviewer so it's clear to everyone?

It's probably worth asking about the rbt issue on Mesos slack to see if others ran into it.

m9a:
The JIRA for this PR: https://issues.apache.org/jira/browse/MESOS-8750
Since @xujyan is shepherding it I intended to set him as the reviewer but it doesn't look like I can change those fields on the PR.

m9a:
@Gilbert88, @xujyan Sure, I will check about the rbt issue on slack.

m9a:
Closing this in favor of https://reviews.apache.org/r/66644/diff/1#index_header

PR #280, opened 2018-04-03
 Fix wrong example of executor message.

carlonelong:
The example of executor message in executor-http-api.md is wrong. This patch fixes this.


xujyan:
Thanks for @carlonelong for the PR. The PR includes some extraneous commits (see above), could you clean it up?

carlonelong:
Sorry, my bad. Check again pls. @xujyan

Gilbert88:
LGTM! I committed this patch with commit message modified.

PR #284, opened 2018-04-17
Remove accent from first name of clems4ever contributor.

clems4ever:
Review bot fails to apply reviews (with ./support/apply-review.py script) because of the accent in my first name.

### DIFF
--- a/docs/contributors.yaml
+++ b/docs/contributors.yaml
@@ -216,7 +216,7 @@
   jira_user: chhsia0
   reviewboard_user: chhsia0
 
-- name: Clément Michaud
+- name: Clement Michaud
   affiliations:
     - {organization: Criteo}
   emails:


clems4ever:
Hello @tillt , there is an issue with the review bot applying my review: https://reviews.apache.org/r/66621/. I think it is due to the accent in my first name.
So, I created this review to remove the accent. Can you please merge it? Thanks.

bmahler:
@clems4ever can you post the error? Is this an issue in `rbt` or our python wrapper script?

clems4ever:
@bmahler , here is the error:

clems4ever:
Apparently the character `\xe9` is `é` as stated here: https://stackoverflow.com/questions/6327494/what-is-the-following-unicode-string-xe9. So, I really think this is due to my first name.

bmahler:
We should try to fix apply-reviews.py. Looks related to this discussion?
https://bugs.python.org/issue1759845

kaysoky:
Here's the associated JIRA: https://issues.apache.org/jira/browse/MESOS-7658

andschwa:
@bmahler Really the only fix is to convert our Python scripts to Python 3. (I tried to get someone at the Hackathon to do this 😉 but failed 😦).

clems4ever:
Thank you guys for your feedbacks. Since it might take some time to rewrite the script in Python 3 I think it's better to merge this PR so that I can move forward with my review in the review board. What do you think?

andschwa:
@clems4ever Yes, and I'm sorry 😭

clems4ever:
@andschwa , no problem. Do you have the rights to merge it?

bbannier:
@clems4ever I don't believe fixing that database is going to make `support/apply_reviews.py` work. I think you need to reconfigure the name you use to create git commit you put onto reviewboard -- this can be done for just the Mesos repo with `git config user name ...`.

Sorry for the inconvenience :/

clems4ever:
@bbannier , apparently apply-review.py retrieves data from the profile in the review board. I updated my profile and now the author is correct (I mean without accent).
Thank you for the help.

PR #285, opened 2018-04-25
mesos.scheduler and mesos.executor seem to cause errors, changing to mesos.native

gouravrattihalli:
A fix for the python framework and executor example.

### DIFF
--- a/src/examples/python/test_executor.py
+++ b/src/examples/python/test_executor.py
@@ -22,7 +22,7 @@ import time
 
 import mesos.interface
 from mesos.interface import mesos_pb2
-from mesos.executor import MesosExecutorDriver
+from mesos.native import MesosExecutorDriver
 
 class MyExecutor(mesos.interface.Executor):
     def launchTask(self, driver, task):
--- a/src/examples/python/test_framework.py
+++ b/src/examples/python/test_framework.py
@@ -22,7 +22,7 @@ import time
 
 import mesos.interface
 from mesos.interface import mesos_pb2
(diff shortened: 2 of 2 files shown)


bmahler:
What were the errors?

gouravrattihalli:
I got an import error.
`ImportError: No module named executor`

Sorry for the super late response.

greggomann:
@tillt would you have cycles to see if this is still an issue?

tillt:
@gouravrattihalli are you sure you were testing the current master and its resulting eggs?

We did split them into two separate eggs `mesos.executor` and `mesos.scheduler` a while ago and adapted the eggs accordingly. It seems to me that you are trying to use older versions of the Mesos python eggs together with more recent example code.

gouravrattihalli:
@tillt I see mesos-executor==1.0.0 on my system but don't see mesos.scheduler. Also looks like the egg is not available on [pypi](https://pypi.org/project/mesos.scheduler/ ).

tillt:
@gouravrattihalli where did you get the "mesos-executor" egg from?

In any case, I would advise you to simply build mesos as a whole from the sources and use the resulting python eggs - they should be complete and the examples should work without modification.

In regards to the pypi repo, I have now created a ticket that tracks any progress on publishing the missing eggs. Please see https://issues.apache.org/jira/browse/MESOS-9195

gouravrattihalli:
@tillt I got the "mesos-executor" egg from [pypi](https://pypi.org/project/mesos-executor/). I was trying to avoid building Mesos, hence used the one from pypi.

PR #289, opened 2018-05-10
Fixtestflags

radhikaj:
Fixing bug described in
https://issues.apache.org/jira/browse/MESOS-8857
echo on windows just reproduces the command line args it receives and does not apply any text processing to it. The Flags test in SubProcessTest expects standard text processing to be applied to the command line args. Therefore instead of using echo on windows to test that flags are being fed correctly to a console app, we add test-flags.exe which receives command line args via argv and prints argv's elements out with a space in between.


andschwa:
Thanks @radhikaj! Closing as we got this in over ReviewBoard and closed MESOS-8857.

PR #290, opened 2018-05-15
Integrated mesos fetcher with libarchive.

johnkord:
After talking to @andschwa , I'm posting a pull request to github due to some issues with binary patches on review board.  I can provide more detail if you're curious. Maybe it's a max patch file size issue, but it cuts the patch off on a line that makes me suspect that it's a bug.

Anyway, here are the review board posts:

https://reviews.apache.org/r/67118/
https://reviews.apache.org/r/67064/
https://reviews.apache.org/r/67065/
https://reviews.apache.org/r/67066/

Since there's no CI on github, I'll need some people to test each build scenario (CMake on Windows, CMake + autotools on Linux). I have verified that all builds work on my systems.


greggomann:
Closing since the associated RRs have landed.

PR #296, opened 2018-06-11
Add xiao wei.

sharego:
Add xiao wei

### DIFF
--- a/docs/contributors.yaml
+++ b/docs/contributors.yaml
@@ -770,6 +770,14 @@
   jira_user: xds2000
   reviewboard_user: xds2000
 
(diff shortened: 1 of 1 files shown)


docs/contributors.yaml
```diff
@@ -825,3 +825,11 @@
     - xingzhou@cn.ibm.com
   jira_user: dongdong
   reviewboard_user: dongdong
+
+- name: xiao wei
```
jieyu: Can you sort based on first name?

sharego: ok, get it

PR #299, opened 2018-06-21
Fix orphaning CNI network on recovery.

vixns:


src/slave/containerizer/mesos/isolators/network/cni/cni.cpp
```diff
@@ -428,6 +428,10 @@ Future<Nothing> NetworkCniIsolatorProcess::recover(
       }
     }
 
+    if (infos.contains(containerId)) {
```
qianzhangxa: This will make the unknown orphaned containers are skipped (i.e., we will miss to do cleanup for unknown orphaned container). So I would suggest to just do a `continue;` between L420 and L421.

@vixns Do you want to post a patch in https://reviews.apache.org/?

src/slave/containerizer/mesos/isolators/network/cni/cni.cpp
```diff
@@ -428,6 +428,10 @@ Future<Nothing> NetworkCniIsolatorProcess::recover(
...
+      continue;
+    }
+
```
qianzhangxa: This will make the unknown orphaned containers are skipped (i.e., we will miss to do cleanup for unknown orphaned container). So I would suggest to just do a continue; between L420 and L421.

@vixns Do you want to post a patch in https://reviews.apache.org/?

vixns: are you sure ? if we move the continue to L421, L423 will have no effect.

qianzhangxa: Why will L423 have no effect? I mean we should just do a `continue;` for the recoverable containers (L410 - L421), for the orphaned containers, we can still keep the current logic.

vixns: Understood, L421 is for orphaned tasks only, and in this case the L437 test matches.
So you're right, moving the continue to L421 make sense.

PR #301, opened 2018-07-10
Document SUPPRESS HTTP call [MESOS-7211]

thzois:


docs/scheduler-http-api.md
```diff
@@ -128,6 +128,26 @@ TEARDOWN Response:
 HTTP/1.1 202 Accepted
 ```
 
+### SUPPRESS
```
vinodkone: Can you move this to #500 below #REQUEST section to preserve the call enum order?

docs/scheduler-http-api.md
```diff
@@ -128,6 +128,26 @@ TEARDOWN Response:
...
+SUPPRESS Request (JSON):
+POST /api/v1/scheduler  HTTP/1.1
+
+Host: masterhost:5050
+Content-Type: application/json
+Mesos-Stream-Id: 130ae4e3-6b13-4ef4-baa9-9f2e85c3e9af
+
+{
+  "framework_id"        : {"value" : "12220-3440-12532-2345"},
+  "type"                        : "SUPPRESS"
```
nfnt: Please re-indent. With spaces, even though some other calls in this document use tabs.

docs/scheduler-http-api.md
```diff
@@ -128,6 +128,26 @@ TEARDOWN Response:
...
+}
```
nfnt: The `Supress` call has a `roles` option. Please use an example call that sets a role to suppress offers for.

docs/scheduler-http-api.md
```diff
@@ -128,6 +128,26 @@ TEARDOWN Response:
 HTTP/1.1 202 Accepted
 ```
 
+### SUPPRESS
+Sent by the scheduler when it wants to inform Mesos master to stop sending offers to the framework. When Mesos receives this request it will stop sending offers to framework but the tasks will continue running for `FrameworkInfo.failover_timeout`. Once the framework has work to do, it can send again a `REVIVE` request.
```
vinodkone: How about:

Sent by the scheduler when it doesn't need offers for a given set of its roles. When Mesos master receives this request, it will stop sending offers for the given set of roles to the framework. As a special case, if `roles` is not specified, all subscribed roles of this framework are suppressed.

Note that master continues to send offers to other subscribed roles of this framework that are not suppressed. Also, status updates about tasks, executors and agents are not affected by this call. 

If the scheduler wishes to receive offers for the suppressed roles again (e.g., it needs to schedule new workloads), it can send `REVIVE` call.

docs/scheduler-http-api.md
vinodkone: Can you also include `roles` field here for completeness?

docs/scheduler-http-api.md
```diff
@@ -128,6 +128,26 @@ TEARDOWN Response:
...
+
+{
+  "framework_id"        : {"value" : "12220-3440-12532-2345"},
+  "type"                        : "SUPPRESS"
+}
+
+SUPPRESS Response:
+HTTP/1.1 202 Accepted
+```
+
```
vinodkone: While you are at it, can you also update the documentation about "REVIVE" call below. I think it's a bit outdated now. Preferably in a different PR.

thzois:
Hello, I think that everything is OK now. Can you confirm this? I will also create a new pull request for the "REVIVE" call.

docs/scheduler-http-api.md
```diff
@@ -479,6 +479,32 @@ HTTP/1.1 202 Accepted
 
 ```
 
+### SUPPRESS
+Sent by the scheduler when it doesn't need offers for a given set of its roles. When Mesos master receives this request, it will stop sending offers for the given set of roles to the framework. As a special case, if roles are not specified, all subscribed roles of this framework are suppressed.
+
+Note that master continues to send offers to other subscribed roles of this framework that are not suppressed. Also, status updates about tasks, executors and agents are not affected by this call and tasks will continue running for `FrameworkInfo.failover_timeout`.
```
vinodkone: I would remove `...and tasks will continue running for `FrameworkInfo.failover_timeout`` part. That timeout doesn't come into play at all during suppression. That will only come into play when a framework disconnects.

docs/scheduler-http-api.md
```diff
@@ -479,6 +479,32 @@ HTTP/1.1 202 Accepted
...
+POST /api/v1/scheduler  HTTP/1.1
+
+Host: masterhost:5050
+Content-Type: application/json
+Mesos-Stream-Id: 130ae4e3-6b13-4ef4-baa9-9f2e85c3e9af
+
+{
+  "framework_id" : {"value" : "12220-3440-12532-2345"},
+  "type"         : "SUPPRESS",
+  "suppress"     : {"role": <one-of-the-subscribed-roles>}
```
vinodkone: This should be `roles` and not `role` since Mesos at least 1.3.0!  And `roles` takes an array of strings (i.e., roles) and not a single role.

PR #305, opened 2018-07-27
Remove redundant sentence from development guide.

GJL:
Remove duplicated sentence from app framework development guide: 
> Your scheduler should create a SchedulerDriver (which will mediate communication between your scheduler and the Mesos master) and then call SchedulerDriver.run():

Preceding sentence is:
> Your scheduler should create a SchedulerDriver (which will mediate communication between your scheduler and the Mesos master) and then call SchedulerDriver.run() (see: C++, Java, Python).

See http://mesos.apache.org/documentation/latest/app-framework-development-guide/#api


jpeach:
Thanks @GJL

PR #306, opened 2018-08-05
Fixed the iptables deadlock in CNI port mapper plugin.

jieyu:
Patch 1: 

It is possible that the port mapping cleanup command will cause iptables
to deadlock if there are a lot of entires in the iptables, because the
`sed` won't process the next line while executing 'iptables -w -t nat -D
...'. But the executing of 'iptables -w -t nat -D ...' might get stuck
if the first command 'iptables -w -t nat -S %s' didn't finish (because
the xtables lock is not released). The first command might not finish if
it has a lot of output, filling the pipe that `sed` hasn't had a chance
to process yet. See more details in MESOS-9127.

This patch fixed the issue by writing the commands to a file and then
executing them.

Patch 2:

Updated port mapper CNI test.

This patch updated the port mapper CNI test to launch multiple
containers concurrently. This would allow us to catch the scenarios
where multiple iptables commands are executed concurrently.

This test fails if the fix for MESOS-9125 is not included.

### DIFF
--- a/src/common/values.hpp
+++ b/src/common/values.hpp
@@ -19,6 +19,7 @@
 
 #include <limits>
 #include <type_traits>
+#include <vector>
 
 #include <mesos/mesos.hpp>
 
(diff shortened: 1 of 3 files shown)


jieyu:
Closed this as the review has been moved to reviewboard.

PR #319, opened 2018-11-18
Fixed incorrect RPM build settings.

thinker0:
Update 'mesos-init-wrapper' execute permission.


jieyu:
@jpeach can you take a look?

PR #323, opened 2018-12-25
Fixed wrong example of --resources=filepath agent configuration.

carlonelong:
Change-Id: I10ac0b5ec20319c1590a63506fbd92f5e82e9dee


carlonelong:
I checked the code and found that --resources=filepath would not work if filepath did not start with "file://".

carlonelong:
@andschwa  would you take a look at this pls?

docs/configuration/agent.md
```diff
@@ -1348,8 +1348,8 @@ As a key:value list:
 <p/>
 To use JSON, pass a JSON-formatted string or use
 <code>--resources=filepath</code> to specify the resources via a file containing
-a JSON-formatted string. 'filepath' can be of the form
-<code>file:///path/to/file</code> or <code>/path/to/file</code>.
+a JSON-formatted string. 'filepath' can only be of the form
```
jieyu: Can you also update here as well?
https://github.com/apache/mesos/blob/master/src/slave/flags.cpp#L76

carlonelong:
@jieyu Updated!

src/slave/flags.cpp
```diff
@@ -74,8 +74,8 @@ mesos::internal::slave::Flags::Flags()
       "\n"
       "To use JSON, pass a JSON-formatted string or use\n"
       "`--resources=filepath` to specify the resources via a file containing\n"
-      "a JSON-formatted string. `filepath` can be of the form\n"
-      "`file:///path/to/file` or `/path/to/file`.\n"
+      "a JSON-formatted string. `filepath` can only be of the form\n"
+      "`file:///path/to/file`.\n"
```
jieyu: in fact, can you test if `file:///` works on the latest Mesos or Mesos 1.7? By reading code, i doubt that `file:///` is going to work ...

carlonelong: Okay... This PR was committed 3 months ago. I'll check the code again.

carlonelong: It works on master(1.9.0).

PR #326, opened 2019-03-11
MESOS-6874: Validate the match between Type and *Infos in the ContainerInfo.

asekretenko:
To avoid situations like described in [MESOS-6874](https://issues.apache.org/jira/browse/MESOS-6874), I'm clarifying the validity criteria for the `ContainerInfo` protobuf and adding validation of this protobuf into the ComposingContainerizer.

There basically has been three ways to validate this protobuf:
1. hardcoded check
2. labelling all the *Infos with protobuf options and use some protobuf reflection
3. require exact match between the Type enum members and the *Info fields, and use a bit more reflection.

Hardcoded check will not "scale well" in case more container types are added. Protobuf options are not a good option in our case (mesos.proto constututes a part of an external interface and adding an option there would require to register it globally). That's why I opted for the third option.

I'm also adding a test for this validation: passing ContainerInfo with Type==MESOS and docker set should make the ComposingContainerizer fail.

After adding this validation I discovered an invalid ContainerInfo used in the test DefaultExecutorWithDockerImageCommandHealthCheck. This PR contains a fix for this test.

Testing done:
- run make check after adding the test only: test expectedly fails
- run make check after adding validation code and fixing the problem with the health check  test: tests pass.

### DIFF
--- a/include/mesos/mesos.proto
+++ b/include/mesos/mesos.proto
@@ -3328,6 +3328,8 @@ message TTYInfo {
  */
 message ContainerInfo {
   // All container implementation types.
+  // For each type there should be a field in the ContainerInfo itself
+  // with exactly matching name in lowercase.
   enum Type {
     DOCKER = 1;
     MESOS = 2;
@@ -3381,8 +3383,8 @@ message ContainerInfo {
   repeated Volume volumes = 2;
   optional string hostname = 4;
 
-  // Only one of the following *Info messages should be set to match
-  // the type.
+  // At most one of the following *Info messages should be set to match
+  // the type, i.e. the "protobuf union" in ContainerInfo should be valid.
   optional DockerInfo docker = 3;
   optional MesosInfo mesos = 5;
 
--- a/src/common/protobuf_utils.cpp
+++ b/src/common/protobuf_utils.cpp
@@ -77,6 +77,62 @@ namespace mesos {
 namespace internal {
 namespace protobuf {
 
+// If `descriptor` is not a descriptor of a protobuf union,
+// this constructor will abort the process.
+UnionValidator::UnionValidator(const google::protobuf::Descriptor* descriptor)
+{
+  const auto* typeFieldDescriptor = descriptor->FindFieldByName("type");
+  CHECK_NOTNULL(typeFieldDescriptor);
+
+  typeDescriptor_ = typeFieldDescriptor->enum_type();
+  CHECK_NOTNULL(typeDescriptor_);
+
+  const auto* unknownTypeValueDescriptor =
+    typeDescriptor_->FindValueByNumber(0);
+  if (unknownTypeValueDescriptor != nullptr) {
+    CHECK_EQ(unknownTypeValueDescriptor->name(), "UNKNOWN");
+  }
+
+  for (int index = 0; index < typeDescriptor_->value_count(); index++) {
+    const auto* typeValueDescriptor = typeDescriptor_->value(index);
+    if (typeValueDescriptor->number() == 0) {
+      // We are skipping the "UNKNOWN" value of the enum.
+      continue;
+    }
+
+    const auto* fieldDescriptor =
+      descriptor->FindFieldByName(strings::lower(typeValueDescriptor->name()));
+
+    CHECK_NOTNULL(fieldDescriptor);
+    unionFieldDescriptors_.emplace_back(
+        typeValueDescriptor->number(), fieldDescriptor);
+  }
+}
+
+
+Option<Error> UnionValidator::validate(
+  const int messageTypeNumber, const google::protobuf::Message& message) const
+{
+  const auto* reflection = message.GetReflection();
+  for (const auto& item : unionFieldDescriptors_) {
+    const auto typeNumber = item.first;
+    const auto* fieldDescriptor = item.second;
+    if (
+        messageTypeNumber != typeNumber &&
+        reflection->HasField(message, fieldDescriptor)) {
+      const auto* descr = typeDescriptor_->FindValueByNumber(messageTypeNumber);
+      return Error(
+          "Protobuf union `" + message.GetDescriptor()->full_name() +
+          "` with `Type == " +
+          (descr == nullptr ? string("<UNKNOWN>") : descr->name()) +
+          "` should not have the field `" +
+          fieldDescriptor->name() + "` set.");
+    }
+  }
+  return None();
+}
+
+
 bool frameworkHasCapability(
     const FrameworkInfo& framework,
     FrameworkInfo::Capability::Type capability)
--- a/src/common/protobuf_utils.hpp
+++ b/src/common/protobuf_utils.hpp
@@ -64,6 +64,51 @@ struct Slave;
 
 namespace protobuf {
 
+// Internal helper class for protobuf union validation.
+class UnionValidator
+{
+public:
+  UnionValidator(const google::protobuf::Descriptor*);
+  Option<Error> validate(
+      const int messageTypeNumber, const google::protobuf::Message&) const;
+
+private:
+  std::vector<std::pair<int, const google::protobuf::FieldDescriptor*>>
+    unionFieldDescriptors_;
+  const google::protobuf::EnumDescriptor* typeDescriptor_;
+};
+
+//
+// A message is a "protobuf union" if, and only if,
+// the following requirements are satisfied:
+// 1. It has a required field named `type` of an enum type.
+// 2. A member of this enum with a number (not index!) of 0
+//    either is named "UNKNOWN" or does not exist.
+// 3. For each other member of this enum there is an optional field
+//    in the message with an exactly matching name in lowercase.
+// (Being or not being a protobuf uinion depends on a message declaration only.)
+//
+// A "protobuf union" is valid if, and only if, all the message fields
+// which correspond to members of this enum that do not matching the value
+// of the `type` field, are not set.
+// (Validity of the protobuf union depends on the message contents.
+// Note that it does not depend on whether the matching field is set or not.)
+//
+// NOTE: If possible, oneof should be used in the new messages instead
+// of the "protobuf union".
+//
+// This function returns None if the protobuf union is valid
+// and Error otherwise.
+// In case the ProtobufUnion is not a protobuf union,
+// this function will abort the process on the first use.
+template <class ProtobufUnion>
+Option<Error> validateProtobufUnion(const ProtobufUnion& message)
+{
+  static const UnionValidator validator(ProtobufUnion::descriptor());
+  return validator.validate(message.type(), message);
+}
+
+
 bool frameworkHasCapability(
     const FrameworkInfo& framework,
     FrameworkInfo::Capability::Type capability);
--- a/src/tests/health_check_tests.cpp
+++ b/src/tests/health_check_tests.cpp
@@ -1819,11 +1819,11 @@ TEST_F_TEMP_DISABLED_ON_WINDOWS(
   }
 }
 
-
+#ifdef __linux__
 // Tests a healthy docker task via CMD health checks using the
 // DefaultExecutor.
 TEST_F_TEMP_DISABLED_ON_WINDOWS(
-    HealthCheckTest, DefaultExecutorWithDockerImageCommandHealthCheck)
+    HealthCheckTest, ROOT_DefaultExecutorWithDockerImageCommandHealthCheck)
 {
   Try<Owned<cluster::Master>> master = StartMaster();
   ASSERT_SOME(master);
@@ -1838,6 +1838,17 @@ TEST_F_TEMP_DISABLED_ON_WINDOWS(
 
   flags.acls = acls;
 #endif // USE_SSL_SOCKET
+
+  const string directory = path::join(os::getcwd(), "archives");
+
+  Future<Nothing> testImage = DockerArchive::create(directory, "alpine");
+  AWAIT_READY(testImage);
+  ASSERT_TRUE(os::exists(path::join(directory, "alpine.tar")));
+
+  flags.isolation = "docker/runtime,filesystem/linux";
+  flags.image_providers = "docker";
+  flags.docker_registry = directory;
+  flags.docker_store_dir = path::join(os::getcwd(), "store");
 
   Fetcher fetcher(flags);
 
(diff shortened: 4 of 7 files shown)


src/slave/containerizer/composing.cpp
```diff
@@ -425,6 +425,28 @@ Future<Containerizer::LaunchResult> ComposingContainerizerProcess::_launch(
 }
 
 
+static Try<Nothing> validateContainerInfoType(const ContainerInfo& info)
```
kaysoky: We use this "protobuf union" trick in tons of different protobufs, like the `CheckInfo`, `HealthCheck`, `ExecutorInfo`, `Value`, `Resource::DiskInfo::Source`, `Offer::Operation`, `Environment::Variable`, `Secret`, `Image`, `Volume::Source`, and `ContainerInfo`.

I wonder if it would be helpful to make this a templated helper function in `src/common/protobuf_utils.hpp`.  Something like this (which I have not tried):
```
template <typename Info>
static Try<Nothing> validateProtobufUnion(const Info& info) 
{
  const auto* typeDescriptor = Info::Type_descriptor();
  const auto* infoDescriptor = Info::descriptor();
  ...
}
```
With usage like:
```
CHECK_SOME(validateProtobufUnion<ContainerInfo>(info));
```

src/slave/containerizer/composing.cpp
```diff
@@ -425,6 +425,28 @@ Future<Containerizer::LaunchResult> ComposingContainerizerProcess::_launch(
...
+  const auto* typeDescriptor = ContainerInfo::Type_descriptor();
+  const auto* infoDescriptor = ContainerInfo::descriptor();
+  for (int index = 0; index < typeDescriptor->value_count(); index++) {
+    const auto* typeValueDescriptor = typeDescriptor->value(index);
+    const auto* infoValueDescriptor = infoDescriptor->FindFieldByName(
+      strings::lower(typeValueDescriptor->name()));
+    CHECK_NOTNULL(infoValueDescriptor);
+
+    if (
+      int(info.type()) != typeValueDescriptor->number() &&
```
kaysoky: It looks like, in every protobuf union except this one, we have an `UNKNOWN = 0` Type, which guards against future additions to the protobuf.  If you end up generalizing this helper, the `0` case will need to be ignored.

include/mesos/mesos.proto
```diff
@@ -3328,6 +3328,9 @@ message TTYInfo {
  */
 message ContainerInfo {
   // All container implementation types.
+  // For each type there should be a field in the ContainerInfo itself
+  // with exactly matching name in the lowercase.
```
kaysoky: I just realized that the protobuf docs removed mention of their union trick a few months ago 😱 
https://github.com/protocolbuffers/protobuf/issues/452

Here is a snapshot of the union trick documentation:
https://web.archive.org/web/20170715111519/https://developers.google.com/protocol-buffers/docs/techniques

kaysoky: Alternatively, we can look at the replacement:
https://developers.google.com/protocol-buffers/docs/proto#oneof

src/tests/containerizer/composing_containerizer_tests.cpp
```diff
@@ -57,6 +57,32 @@ namespace tests {
 
 class ComposingContainerizerTest : public MesosTest {};
 
+TEST_F(ComposingContainerizerTest, FailsOnContainerInfoTypeMismatch)
```
kaysoky: It should be possible to have the same validation error when only using a single containerizer.  

i.e. When the `DockerInfo` is filled in with a `Mesos` Type, that is treated like an empty `MesosInfo` instead of an error.  This is true even if the Mesos containerizer is the only active containerizer.

As such, we'd ideally have a parameterized test for each of the three possible cases (`mesos`, `mesos,docker`, and `docker`).  Here's a test that is parameterized by the `--containerizers` agent flag:
https://github.com/apache/mesos/blob/master/src/tests/default_executor_tests.cpp#L111-L124

lava: I'm not sure that I understand your comment - the validation acts on the protobuf message, so the value of the `--containerizers` flag should not make any difference?

src/tests/health_check_tests.cpp
```diff
@@ -1889,9 +1900,12 @@ TEST_F_TEMP_DISABLED_ON_WINDOWS(
   TaskInfo task = createTask(offers->front(), "sleep 120");
 
   // TODO(tnachen): Use local image to test if possible.
+  Image image;
+  image.set_type(Image::DOCKER);
+  image.mutable_docker()->set_name("alpine");
```
lava: I'd suggest adding a blank line here.

include/mesos/mesos.proto
lava: We looked at `oneof` while writing the patches, but we cannot use it here because it breaks backwards compatibility.

asekretenko: Unfortunately, there are two problems with converting this external interface to `oneof`.
- First, it would make the code that sets both `docker` and `mesos` silently do strange things (instead of failing). I cannot prove that there is no such code outside of Mesos.
- Second, we will still need `ContainerInfo::Type` and thus the validation. There are a lot of tests that only do `set_type(Type::MESOS)` but set neither `docker` nor `mesos`. According to the comment above the `ContainerInfo` definition, this is a perfectly valid way to set up a `MESOS` container without a filesystem isolation. I would expect such code to also exist in production outside of Mesos itself.

It would have been a right decision to use `oneof` when this interface was initially introduced (and do without the `type` field), but now it is too late:(

src/tests/containerizer/composing_containerizer_tests.cpp
asekretenko: Actually, this test produces a validation error without feeding _any_ real containerizer into the `ComposingContainerizer`. Only the mock containerizer is used.

src/slave/containerizer/composing.cpp
asekretenko: Making this a templated helper makes sense, I'll do that.

Using this helper for all the other similar protobufs looks like another task of an unknown difficulty and volume;)

src/tests/containerizer/composing_containerizer_tests.cpp
kaysoky: I'm referring to the how the containerizers are passed messages.  When the `--containerizers` flag only has a single entry (`mesos` or `docker`), the `ComposingContainerizer` is not used.  That means the validation logic added here will not be run, even though the validation logic is still useful in these cases.

src/slave/containerizer/composing.cpp
asekretenko: Converted this into a templated helper.

src/tests/containerizer/composing_containerizer_tests.cpp
asekretenko: After discussion with Benno I moved the validation to master.

src/common/protobuf_utils.hpp
```diff
@@ -64,6 +64,63 @@ struct Slave;
...
+};
+
+
+template <class Message>
+class UnionValidator
+{
+public:
+  static Option<Error> Validate(const Message& message)
+  {
+    static const UnionValidatorHelper helper(
```
lava: Couldn't you just move this static variable into `validateProtobufUnion()` and get rid of this class? If yes, we could also rename `UnionValidatorHelper` -> `UnionValidator`.

src/common/protobuf_utils.hpp
```diff
@@ -64,6 +64,63 @@ struct Slave;
...
+        Message::descriptor(), Message::Type_descriptor());
+    return helper.Validate(message.type(), message);
+  };
+  UnionValidator() = delete;
+};
+
+
+// This function returns Error if the protobuf union is not valid
+// and Nothing otherwise.
```
lava: Let's also mention that it will abort the process if the passed message is not a protobuf union. (which is actually the correct behaviour, I think, because getting the wrong message type here is essentially a compile-time bug that is only detected at runtime)

src/common/protobuf_utils.hpp
```diff
@@ -64,6 +64,63 @@ struct Slave;
...
+// of the "protobuf union".
+//
+class UnionValidatorHelper
+{
+public:
+  UnionValidatorHelper(
+      const google::protobuf::Descriptor*,
+      const google::protobuf::EnumDescriptor*);
+
+  Option<Error> Validate(
```
lava: Member function names and member field names should start with a lowercase letter. Also, not required but highly recommended is to use trailing underscores for field names. (http://mesos.apache.org/documentation/latest/c++-style-guide/#naming)

asekretenko:
Test results from CI:

Configuration Name | Duration | All | Failed | Skipped
-- | -- | -- | -- | --
FLAG=Plain,label=mesos-ec2-ubuntu-16.04 | 18 min | 3312 | 0 | 0
FLAG=BUILD_ISOLATORS,label=mesos-ec2-ubuntu-16.04 | 5.4 sec | 607 | 0 | 0
FLAG=CMake,label=mesos-ec2-centos-7 | 22 min | 3256 | 0 | 0
FLAG=SSL,label=mesos-ec2-centos-6 | 23 sec | 677 | 1 | 0
FLAG=SSL,label=mesos-ec2-debian-9 | 16 min | 3297 | 0 | 0
FLAG=CMake,label=mesos-ec2-ubuntu-16.04 | 19 min | 3265 | 0 | 0
FLAG=Clang,label=mesos-ec2-ubuntu-16.04 | 18 min | 3312 | 0 | 0
FLAG=SSL,label=mesos-ec2-ubuntu-16.04 | 18 min | 3383 | 0 | 0
FLAG=SSL,label=mesos-ec2-centos-7 | 20 min | 3381 | 0 | 0
FLAG=SSL,label=mesos-ec2-debian-8 | 20 min | 3379 | 0 | 0
FLAG=SSL,label=mesos-ec2-ubuntu-14.04 | 19 min | 3382 | 0 | 0
FLAG=SSL,label=mac | 9 min 19 sec | 2775 | 0 | 0
FLAG=Plain,label=mesos-ec2-centos-7 | 20 min | 3303 | 0 | 0

centos-7 seems to have failed due to some completely unrelated problem:

> 19:22:48 ../../3rdparty/libprocess/include/process/gmock.hpp:247: ERROR: this mock object (used in test SlaveTest.MetricsSlaveLaunchErrors) should be deleted but never is. Its address is @0x56279ea978e8.
19:22:48 ../../src/tests/mock_registrar.cpp:54: ERROR: this mock object (used in test SlaveTest.MetricsSlaveLaunchErrors) should be deleted but never is. Its address is @0x56279eb0d9f0.
19:22:48 ../../src/tests/containerizer.cpp:387: ERROR: this mock object (used in test SlaveTest.MetricsSlaveLaunchErrors) should be deleted but never is. Its address is @0x7ffd372b1220.
19:22:48 ../../src/tests/slave_tests.cpp:1400: ERROR: this mock object (used in test SlaveTest.MetricsSlaveLaunchErrors) should be deleted but never is. Its address is @0x7ffd372b16c0.

src/common/protobuf_utils.hpp
asekretenko: Exactly. We do not need this class at all. Moved and renamed, as you suggested.

src/common/protobuf_utils.cpp
```diff
@@ -77,6 +77,56 @@ namespace mesos {
 namespace internal {
 namespace protobuf {
 
+
+UnionValidator::UnionValidator(
+  const google::protobuf::Descriptor* descriptor,
+  const google::protobuf::EnumDescriptor* typeDescriptor)
+{
+  typeDescriptor_ = typeDescriptor;
+  const auto* unknownTypeValueDescriptor = typeDescriptor->FindValueByNumber(0);
```
kaysoky: Consider leaving a newline after `typeDescriptor_ = typeDescriptor;` and inserting a short comment after this section, i.e.:
```
// If present, the the `0` value of the `Type` enum must be named "UNKNOWN".
const auto* unknownTypeValueDescriptor = typeDescriptor->FindValueByNumber(0);
...
```

src/common/protobuf_utils.hpp
```diff
@@ -64,6 +64,53 @@ struct Slave;
...
+  const google::protobuf::EnumDescriptor* typeDescriptor_;
+};
+
+
+// This function returns Error if the protobuf union is not valid
+// and Nothing otherwise.
+// In case MessageWithUnion is not a protobuf union,
+// this function will abort the process on the first use.
+template <class MessageWithUnion>
+Option<Error> validateProtobufUnion(const MessageWithUnion& message)
```
kaysoky: This might be easy to miss from reading the implementation:
```
// NOTE: A protobuf union is only invalid when unexpected member fields are set.
// It is valid when all members of a protobuf union are empty.
```

src/common/protobuf_utils.hpp
```diff
@@ -64,6 +64,53 @@ struct Slave;
...
+{
+  static const UnionValidator validator(
+      MessageWithUnion::descriptor(), MessageWithUnion::Type_descriptor());
```
kaysoky: Looking at this again, arriving at a `google::protobuf::EnumDescriptor` for the second argument can be easily missed, especially for those unfamiliar with protobuf generated code.  A note like this might help:
```
// NOTE: The `::Type_descriptor()` method is a protobuf generated method corresponding
// to the `Type` field of the given protobuf.  We expect this to be an enum field.
```

asekretenko: I've just realized that the requirement that the field `type` has the type `Type` is 
a) not checked anywhere 
and 
b) is completely useless now, when the UnionValidator is initialized once per message type.
I'll simply remove this argument.

asekretenko: I moved the enum descriptor extraction into the UnionValidator contructor. Now it is much more local, and I hope no comment is needed to understand what is going on there.

src/common/protobuf_utils.hpp
asekretenko: I rewrote the comment which explains what a "protobuf union" is almost completely. I hope now this point is more difficult to miss.

src/common/protobuf_utils.cpp
asekretenko: Now it should be much more clear from the code of the UnionValidator constructor to which field the typeDescriptor corresponds.

PR #332, opened 2019-04-11
Fix detection of ELFIO header

Digital-Chaos:
Build fails to detect system provided ELFIO headers when using `configure --with-elfio` flag. Basically this is the same fix as the patch provided here: [https://issues.apache.org/jira/browse/MESOS-6605](https://issues.apache.org/jira/browse/MESOS-6605)


bmahler:
@vinodkone it's probably the first time anyone has made use of the flag to use an external elfio lib

lava:
Thanks for contributing!

PR #334, opened 2019-04-30
Remove u8 prefix to fit C++20

QuellaZhang:
Related to https://issues.apache.org/jira/browse/MESOS-9718.


kaysoky:
Building on @rukletsov 's comment in the JIRA, this change reverts a previous commit (https://reviews.apache.org/r/58430/), which will break the Windows build on prior (released) MSVCs.

cf-natali:
Stale, and unlikely Mesos will be updated to C++20...

PR #335, opened 2019-05-19
Return early in fetcher if no uris are specified.

carlonelong:
Change-Id: I9e78b0a9a58e6d84885eae768a5d76e13288c7af

### DIFF
--- a/src/slave/containerizer/fetcher.cpp
+++ b/src/slave/containerizer/fetcher.cpp
@@ -366,6 +366,9 @@ Future<Nothing> FetcherProcess::fetch(
     const string& sandboxDirectory,
     const Option<string>& user)
 {
+  if (commandInfo.uris_size() == 0) {
+    return Nothing();
+  }
   VLOG(1) << "Starting to fetch URIs for container: " << containerId
           << ", directory: " << sandboxDirectory;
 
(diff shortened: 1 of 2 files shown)


lava:
Hey @carlonelong , 

thanks for contributing!

Your patch looks reasonable to me, however when I ran it through our CI it seemed to break a few unit tests, in particular:

Test Name | Duration | Age
-- | -- | --
mesos-ec2-ubuntu-16.04-SSL.Mesos.DockerContainerizerTest.ROOT_DOCKER_UNPRIVILEGED_USER_NonRootSandbox | 6 sec | 1
mesos-ec2-ubuntu-16.04-SSL.Mesos.MesosContainerizerExecuteTest.ROOT_UNPRIVILEGED_USER_SandboxFileOwnership | 0.1 sec | 1
mesos-ec2-ubuntu-16.04-SSL.Mesos.MesosContainerizerDestroyTest.DestroyWhileFetching | 15 sec | 1
mesos-ec2-ubuntu-16.04-SSL.Mesos.ProvisionerDockerTest.ROOT_INTER

(and the same for debian9, centos6/7, ubuntu 14.04, etc.)

I did not have the time to investigate what exactly is causing these tests to fail, but I think this will need to be investigated before we can commit the patch.

Best regards,
Benno

carlonelong:
@lava 
Hi. After reading fetcher.cpp more carefully, I find that the reason these tests failed is that we'll call "chown" in FetcherProcess::run. We may move this part ahead to solve this problem. But I think it's not elegant. 

I also checked MESOS-5218/6843/6217. And I agreed with what Jie Yu said:"The fetcher code is a mess. We should try to clean it up if possible IMO."

FIFO and named pipes used in containerd(between shim and its container) might be good examples.

PR #341, opened 2019-07-02
Remove unnecessary logging.

bgalek:
There are metrics added in https://github.com/apache/mesos/commit/d93906a so this logging is not needed anymore.

### DIFF
--- a/src/slave/slave.cpp
+++ b/src/slave/slave.cpp
@@ -7157,11 +7157,6 @@ void Slave::_checkDiskUsage(const Future<double>& usage)
     LOG(ERROR) << "Failed to get disk usage: "
                << (usage.isFailed() ? usage.failure() : "future discarded");
   } else {
-    executorDirectoryMaxAllowedAge = age(usage.get());
-    LOG(INFO) << "Current disk usage " << std::setiosflags(std::ios::fixed)
-              << std::setprecision(2) << 100 * usage.get() << "%."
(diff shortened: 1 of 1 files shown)


janisz:
@tillt @bmahler @rukletsov Can you take a look at this?

src/slave/slave.cpp
```diff
@@ -7157,11 +7157,6 @@ void Slave::_checkDiskUsage(const Future<double>& usage)
     LOG(ERROR) << "Failed to get disk usage: "
                << (usage.isFailed() ? usage.failure() : "future discarded");
   } else {
-    executorDirectoryMaxAllowedAge = age(usage.get());
```
lava: Here `executorDirectoryMaxAllowedAge` is actually a member variable of `Slave` that is used for generating the value of the metric, so removing the assignment will probably break this. (Imho this is also a fault in our C++ style guide, having no visual indication of member variables makes it too easy to make this kind of mistake)

vinodkone:
This log line has been valuable to see disk usage over time from logs when we encountered disk full issues. A metric is not a substitute because we would need a metrics stack with history to get the same data. Also, this is the only log line that I know of in the agent that periodically prints something which is valuable because one can easily tell from the logs whether the main actor was stuck or not when debugging issues. 

@bgalek Has this logging been causing any issues for you in production?

bgalek:
@vinodkone nope, just thought it was duplication of metric

PR #353, opened 2020-03-19
Hook manager mutex and global variable enhancement.

bluezd:
This patch aims to fix issue [MESOS-9337]
  1. add missing mutex acquisition for the potential race
  2. change the global mutex and map variable to POD type

Signed-off-by: Dong Zhu <dzhu@d2iq.com>

### DIFF
--- a/src/hook/manager.cpp
+++ b/src/hook/manager.cpp
@@ -47,8 +47,9 @@ using mesos::modules::ModuleManager;
 namespace mesos {
 namespace internal {
 
-static std::mutex mutex;
-static LinkedHashMap<string, Hook*> availableHooks;
+static std::mutex* mutex = new std::mutex();
+static LinkedHashMap<string, Hook*>* availableHooks =
+  new LinkedHashMap<string, Hook*>();
 
 
 Try<Nothing> HookManager::initialize(const string& hookList)
@@ -56,7 +57,7 @@ Try<Nothing> HookManager::initialize(const string& hookList)
   synchronized (mutex) {
     const vector<string> hooks = strings::split(hookList, ",");
     foreach (const string& hook, hooks) {
-      if (availableHooks.contains(hook)) {
+      if (availableHooks->contains(hook)) {
         return Error("Hook module '" + hook + "' already loaded");
       }
 
@@ -73,7 +74,7 @@ Try<Nothing> HookManager::initialize(const string& hookList)
       }
 
       // Add the hook module to the list of available hooks.
-      availableHooks[hook] = module.get();
+      (*availableHooks)[hook] = module.get();
     }
   }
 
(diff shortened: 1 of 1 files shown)


src/hook/manager.cpp
```diff
@@ -240,18 +242,20 @@ Future<DockerTaskExecutorPrepareInfo>
...
-    futures.push_back(
-        hook->slavePreLaunchDockerTaskExecutorDecorator(
-            taskInfo,
-            executorInfo,
-            containerName,
-            containerWorkDirectory,
-            mappedSandboxDirectory,
-            env));
+  synchronized (mutex) {
+    futures.reserve((*availableHooks).size());
```
bmahler: How about `availableHooks->size()` here and all other `(*availableHooks).method()` cases in this patch? That would be more idiomatic with our code base

src/hook/manager.cpp
```diff
@@ -47,16 +47,16 @@ using mesos::modules::ModuleManager;
 namespace mesos {
 namespace internal {
 
-static std::mutex mutex;
-static LinkedHashMap<string, Hook*> availableHooks;
+static std::mutex *mutex = new(std::mutex);
+static LinkedHashMap<string, Hook*> *availableHooks = new(LinkedHashMap<string, Hook*>);
```
bmahler: We place pointers next to the type:

```
static std::mutex* mutex ...
static LinkedHashMap<string, Hook*>* availableHooks
```

bluezd:
@bmahler Thanks for your info, I will update the patch on my own and submit later.

src/hook/manager.cpp
bluezd: Agree with it.

src/hook/manager.cpp
bluezd: Roger that.

bluezd:
@bmahler I have re-submitted the changes. Could you please review it again ?

bmahler:
@asekretenko these are good questions to be asking, as I also felt this looked a bit strange. I suspect it's just modeled in the same way that ModuleManager is written which has quite a bit more to it but is also an all static class with similar static non-POD issues. For unload, it looks like it is used for testing so I'm not sure how easily it could be removed. At this point I haven't read through all the module related code enough to be contemplating changing the design, so I can't write a useful TODO or ticket (but please feel free to if you're interested in pitching in to clean this up!)

Since @bluezd is a new contributor and was interested in hooks/modules, I looked for some small tickets for him to get some experience with the contribution process, and so this patch is just focusing on the trivial work of removing static non-PODs (which we ban) and fixing some unsafe locking in this code.

src/hook/manager.cpp
```diff
@@ -47,16 +47,16 @@ using mesos::modules::ModuleManager;
 namespace mesos {
 namespace internal {
 
-static std::mutex mutex;
-static LinkedHashMap<string, Hook*> availableHooks;
+static std::mutex* mutex = new(std::mutex);
+static LinkedHashMap<string, Hook*>* availableHooks = new(LinkedHashMap<string, Hook*>);
```
bmahler: I'm not familiar with this `new` syntax, we would use:

```
static std::mutex* mutex = new std::mutex();
static LinkedHashMap<string, Hook*>* availableHooks = new LinkedHashMap<string, Hook*>();
```

bmahler: This line is also > 80 characters. Are you using the commit hook? It would catch this issue:

http://mesos.apache.org/documentation/latest/advanced-contribution/#create-your-patch

bluezd: Thanks for your info, just installed the commit hook and make changes to the code accordingly per your suggestion.

PR #354, opened 2020-03-23
Avoid error on mesos logs upon docker image fetch

kamaradclimber:
When fetching docker image, fetcher test curl support for --http1.1
flag.
This patch makes sure we properly capture stderr to avoid displaying
warnings in mesos logs.


bmahler:
Thanks @kamaradclimber, we chatted a bit in slack about this since it looked a bit surprising to me:

https://mesos.slack.com/archives/C1YN31DPA/p1584988424011600

One of us will get to this shortly.

PR #357, opened 2020-04-13
Fixed missing find_if namespace qualification

Bo98:
See 0273fbc2638c581528cc5819da2e0c9b1a9c235e for why this is needed. This one was just left out from that commit, but is included in FreeBSD's patches.


bmahler:
Thanks! This is now committed

PR #358, opened 2020-04-15
Mention that tests should be run as root/admin in contribution guides.

cf-natali:
@abudnik


bmahler:
@abudnik perhaps we could be a bit more specific about when to run as root?

We have a lot of test filters so it's a little more generalized (e.g. if you're touching a 'curl' related test make sure that runs, if touching a docker related test make sure that runs, etc): https://github.com/apache/mesos/blob/master/src/tests/environment.cpp

I'm not sure we'll be able to write a good guideline here, other than just generally explaining how tests can be filtered for various reasons and then relying on contributors / reviewers / CI to make sure the appropriate ones are run.

PR #360, opened 2020-04-27
Fix java bindings in cmake build.

akornatskyy:


akornatskyy:
@asekretenko @greggomann Fixed mesos java bindings for cmake by reversing dependency so `libmesos-java` depends on `libmesos`.

PR #364, opened 2020-05-17
MESOS-10129 Build fails on Maven javadoc generation when using JDK11

csaltos:
Maven Javadoc plugin tries to include non-existant java6 documentation references and it fails ... this fix tries to skip those invalid references for the Javadoc plugin to success (it also works for Java8 and Java11)

### DIFF
--- a/src/java/mesos.pom.in
+++ b/src/java/mesos.pom.in
@@ -86,6 +86,7 @@
         <configuration>
           <sourcepath>@abs_top_srcdir@/src/java/src:@abs_top_builddir@/src/java/generated</sourcepath>
           <subpackages>org.apache.mesos</subpackages>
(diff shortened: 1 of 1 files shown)


csaltos:
The issue at Jira -> https://issues.apache.org/jira/browse/MESOS-10129

cf-natali:
It seems to be actually due to an openjdk bug which has been fixed and backported:
https://bugs.openjdk.java.net/browse/JDK-8212233

@surahman confirmed that it works fine on a recent Ubuntu.

@qianzhangxa can you close this PR?

qianzhangxa:
Sure, thanks @cf-natali and @surahman!

qianzhangxa:
Oh, actually I do not have the permission to manually close PRs in GitHub, I will ask for that.

PR #365, opened 2020-05-17
Fix FreeBSD memory detection (MESOS-10104)

Digital-Chaos:
Fixed detection of available free memory and swap.
Tested on FreeBSD-12.1-STABLE (sysutils/apache-mesos port)

https://issues.apache.org/jira/browse/MESOS-10104

### DIFF
--- a/3rdparty/stout/include/stout/os/freebsd.hpp
+++ b/3rdparty/stout/include/stout/os/freebsd.hpp
@@ -88,17 +88,28 @@ inline Try<Memory> memory()
   const size_t pageSize = os::pagesize();
 
   unsigned int freeCount;
-  size_t length = sizeof(freeCount);
-
+  size_t freeCountLength = sizeof(freeCount);
   if (sysctlbyname(
-      "vm.stats.v_free_count",
+      "vm.stats.vm.v_free_count",
       &freeCount,
-      &length,
+      &freeCountLength,
       nullptr,
       0) != 0) {
     return ErrnoError();
   }
-  memory.free = Bytes(freeCount * pageSize);
+
+  unsigned int inactiveCount;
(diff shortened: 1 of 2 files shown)


bmahler:
@Digital-Chaos thanks for fixing this!

Whoops, looks like we don't have any unit tests for os::memory()! Could you add some here to ensure this fix is covered by tests:
 https://github.com/apache/mesos/blob/1.9.0/3rdparty/stout/tests/os_tests.cpp

Digital-Chaos:
Added a basic test for os::memory as requested. I wasn't sure how to verify the summation of the invalid pages to the free pages, but at least there is more coverage than before!  :-)

bmahler:
@Digital-Chaos Thanks! Looked through the sysctl related documentation and the code makes sense to me. I will ship this as is.

However, I wonder why the code bothers with the lookup by name -> mib transformation and doesn't just do by name (slower) or by mib (faster) entirely:

```
sysctlbyname("vm.swap_info.0", ...)
sysctlbyname("vm.swap_info.1", ...)
// until ENOENT

// Or:
int mib[3] = { CTL_VM, XXX, 0 }; // Oh.. I can't find something that looks like swap_info,
                                 // is there not a mib constant we can use for every name we can lookup?
sysctl(...);
```

(This applies to the earlier sysctlbyname calls in this function as well). Is it because not all entries have mib constants that we can just fill in without lookup?

Digital-Chaos:
> Is it because not all entries have mib constants that we can just fill in without lookup?

Yes, I believe that is the case.

The other way I have seen this done is to call sysctlnametomib() once in some sort of initialisation and hold the mib somewhere to use later in any calls to sysctl(). 
e.g. https://github.com/hishamhm/htop/blob/master/freebsd/FreeBSDProcessList.c

PR #367, opened 2020-09-10
Expose hierarchical allocator recovery parameters as master flags

Lqp1:
Hello,

We needed to expose two parameters of the hierarchical allocator recovery options as master flags. Are you interested in such a PR?

We needed to tweak those values, and rebuilding mesos + deploy it again each time was not very time saving. These flags can be used to have per DC parameters also.

It's rather simple but if you have any input on how I can improve it, don't hesitate.

Br

### DIFF
--- a/include/mesos/allocator/allocator.hpp
+++ b/include/mesos/allocator/allocator.hpp
@@ -70,8 +70,8 @@ struct Options
   bool publishPerFrameworkMetrics = true;
 
   // Recovery options
-  Duration recoveryTimeout = mesos::internal::master::DEFAULT_RECOVERY_TIMEOUT;
-  float agentRecoveryFactor = mesos::internal::master::DEFAULT_AGENT_RECOVERY_FACTOR;
+  Duration recoveryTimeout = mesos::internal::master::DEFAULT_ALLOCATOR_RECOVERY_TIMEOUT;
+  double agentRecoveryFactor = mesos::internal::master::DEFAULT_ALLOCATOR_AGENT_RECOVERY_FACTOR;
 };
 
 
--- a/include/mesos/allocator/allocator.hpp
+++ b/include/mesos/allocator/allocator.hpp
@@ -28,6 +28,8 @@
 
 #include <mesos/quota/quota.hpp>
 
+#include <master/constants.hpp>
+
 #include <mesos/resource_quantities.hpp>
 #include <mesos/resources.hpp>
 
@@ -66,6 +68,10 @@ struct Options
   size_t maxCompletedFrameworks = 0;
 
   bool publishPerFrameworkMetrics = true;
+
+  // Recovery options
+  Duration recoveryTimeout = mesos::internal::master::DEFAULT_RECOVERY_TIMEOUT;
+  float agentRecoveryFactor = mesos::internal::master::DEFAULT_AGENT_RECOVERY_FACTOR;
 };
 
 
@@ -120,7 +126,6 @@ private:
 
   OfferConstraintsFilter(internal::OfferConstraintsFilterImpl&& impl_);
 };
-
 
 /**
  * Per-framework allocator-specific options that are not part of
--- a/include/mesos/allocator/allocator.hpp
+++ b/include/mesos/allocator/allocator.hpp
@@ -28,8 +28,6 @@
 
 #include <mesos/quota/quota.hpp>
 
-#include <master/constants.hpp>
-
 #include <mesos/resource_quantities.hpp>
 #include <mesos/resources.hpp>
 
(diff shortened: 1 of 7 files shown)


include/mesos/allocator/allocator.hpp
```diff
@@ -28,6 +28,8 @@
 
 #include <mesos/quota/quota.hpp>
 
+#include <master/constants.hpp>
```
asekretenko, line 31: The definitions in the public allocator interface must not depend on the internal ones. 
I.e. the headers in `include/` should be usable without ones from `src/`.

Is it possible to define these constants here, either in this header or in a some other header (in the worst case, in a new one)?

src/master/allocator/mesos/hierarchical.hpp
```diff
@@ -755,6 +755,8 @@ class HierarchicalAllocatorProcess : public MesosAllocatorProcess
 
   // Recovery data.
   Option<int> expectedAgentCount;
+  float agentRecoveryFactor;
```
asekretenko, line 758: Is it possible to just use these values from `options`? Why keep a copy?

src/master/flags.cpp
```diff
@@ -460,6 +460,18 @@ mesos::internal::master::Flags::Flags()
...
       DEFAULT_ALLOCATOR);
 
+  add(&Flags::hierarchical_recovery_factor,
+      "hierarchical_recovery_factor",
+      "Ratio of minimal re-registred agent before sending\n"
+      "offers after a leader re-election",
+      DEFAULT_AGENT_RECOVERY_FACTOR);
+
+  add(&Flags::hierarchical_recovery_timeout,
+      "hierarchical_recovery_timeout",
```
asekretenko, line 470: Maybe `allocator_recovery_timeout`? 

In the master code that fills `allocator::Options` nothing implies that the created allocator is indeed _the_ hierarchical allocator. 
Moreover, the existence of recovery interval is not relevant to the fact that the allocator treats resource roles hierarchically.

src/master/flags.cpp
```diff
@@ -460,6 +460,18 @@ mesos::internal::master::Flags::Flags()
       "load an alternate allocator module using `--modules`.",
       DEFAULT_ALLOCATOR);
 
+  add(&Flags::hierarchical_recovery_factor,
+      "hierarchical_recovery_factor",
```
asekretenko, line 464: `allocator_agent_recovery_factor`? (see my comment on the word  "hierarchical" below)

src/master/flags.cpp
```diff
@@ -460,6 +460,18 @@ mesos::internal::master::Flags::Flags()
...
+      "Ratio of minimal re-registred agent before sending\n"
+      "offers after a leader re-election",
+      DEFAULT_AGENT_RECOVERY_FACTOR);
```
asekretenko, line 467: `DEFAULT_ALLOCATOR_AGENT_RECOVERY_FACTOR`? 
Looks like the already existing `Flags` code follows the convention that the default has the same name as the flag, prefixed with `DEFAULT_`?

src/master/flags.cpp
```diff
@@ -460,6 +460,18 @@ mesos::internal::master::Flags::Flags()
       "load an alternate allocator module using `--modules`.",
       DEFAULT_ALLOCATOR);
 
+  add(&Flags::hierarchical_recovery_factor,
+      "hierarchical_recovery_factor",
+      "Ratio of minimal re-registred agent before sending\n"
```
asekretenko, line 465: "Ratio" usually implies two numbers (ratio of something to something else), doesn't it?..

Maybe something  like "Minimum fraction of known agents re-registered after leader election for the allocator to start generating offers"?

src/master/flags.hpp
```diff
@@ -81,6 +81,8 @@ class Flags : public virtual logging::Flags
   Option<std::string> modulesDir;
   std::string authenticators;
   std::string allocator;
+  float hierarchical_recovery_factor;
```
asekretenko, line 84: I would suggest leaving `double`, as Mesos code generally uses `double` and `double` is slightly more difficult to use incorrectly.

Note that `0.80` in the line `constexpr float DEFAULT_AGENT_RECOVERY_FACTOR = 0.80;` is a **double** literal, not a float.

asekretenko (commented):
Thanks for addressing this TODO; might be useful to other users running large clusters.

Overall looks good; there is a number of minor issues to be addressed before merging this (see below).

src/master/flags.cpp
Lqp1: You're right; I used your sentence it seems fine to me too.

src/master/allocator/mesos/hierarchical.hpp
Lqp1: My bad I reworked my patch and forgot to clean this.

include/mesos/allocator/allocator.hpp
Lqp1: So the flags defined for master will depend on `include/allocator/allocator.hpp` header (instead of the more generic `constants.hpp`? Is that ok?

asekretenko (approved):
Thanks! 
Looks good.
I'll merge this, looks like my recent additions to options cause a conflict.

Lqp1:
Thanks!

PR #383, opened 2021-05-02
Add mesos authentication to the mesos cli

andreaspeters:
The following points I done:

- Add authentication against mesos manager and agent
- Add option to skip SSL verification of the mesos-agent
- Changed the order of "task list" to get back more running states

### DIFF
--- a/src/python/cli_new/lib/cli/config.py
+++ b/src/python/cli_new/lib/cli/config.py
@@ -119,6 +119,18 @@ class Config():
 
         return master
 
+    def principal(self):
+        """
+        Return the principal in the configuration file
+        """
+        return self.data["master"]["principal"]
+
+    def secret(self):
+        """
+        Return the secret in the configuration file
+        """
+        return self.data["master"]["secret"]
+
     def plugins(self):
         """
         Parse the plugins listed in the configuration file and return them.
--- a/src/python/cli_new/lib/cli/config.py
+++ b/src/python/cli_new/lib/cli/config.py
@@ -21,6 +21,7 @@ Config class to manage the configuration file.
 import os
 import toml
 
+import requests
 import cli
 
 from cli.constants import DEFAULT_MASTER_IP
@@ -123,13 +124,74 @@ class Config():
         """
         Return the principal in the configuration file
         """
-        return self.data["master"]["principal"]
+        return self.data["master"].get("principal")
 
     def secret(self):
         """
         Return the secret in the configuration file
         """
-        return self.data["master"]["secret"]
+        return self.data["master"].get("secret")
+
+    def agent_ssl(self, default=False):
+        """
+        Return if the agent support ssl
+        """
+        if "agent" in self.data:
+            agent_ssl = self.data["agent"].get("ssl", default)
+            if not isinstance(agent_ssl, bool):
+                raise CLIException("The 'agent->ssl' field"
+                                   " must be True/False")
+
+            return agent_ssl
+
+        return default
+
+    def agent_ssl_verify(self, default=False):
+        """
+        Return if the ssl certificate should be verified
+        """
+        if "agent" in self.data:
+            ssl_verify = self.data["agent"].get("ssl_verify", default)
+            if not isinstance(ssl_verify, bool):
+                raise CLIException("The 'agent->ssl_verify' field"
+                                   " must be True/False")
+
+            return ssl_verify
+
+        return default
+
+    def agent_timeout(self, default=5):
+        """
+        Return the connection timeout of the agent
+        """
+        if "agent" in self.data:
+            timeout = self.data["agent"].get("timeout", default)
+            if not isinstance(timeout, int):
+                raise CLIException("The 'agent->timeout' field"
+                                   " must be a number in seconds")
+
+            return timeout
+
+        return default
+
+
+    def agent_principal(self):
+        """
+        Return the principal in the configuration file
+        """
+        if "agent" in self.data:
+            return self.data["agent"].get("principal")
+
+        return None
+
+    def agent_secret(self):
+        """
+        Return the secret in the configuration file
+        """
+        if "agent" in self.data:
+            return self.data["agent"].get("secret")
+
+        return None
 
     def plugins(self):
         """
@@ -149,3 +211,15 @@ class Config():
             return self.data["plugins"]
 
         return []
+
+    def authentication_header(self):
+        """
+        Return the BasicAuth authentication header
+        """
+        if (self.agent_principal() is not None
+                and self.agent_secret() is not None):
+            return requests.auth.HTTPBasicAuth(
+                self.agent_principal(),
+                self.agent_secret()
+            )
+        return None
--- a/src/python/cli_new/lib/cli/http.py
+++ b/src/python/cli_new/lib/cli/http.py
@@ -19,17 +19,16 @@ A collection of http related functions used by the CLI and its Plugins.
 """
 
 import json
-import urllib.request
-import urllib.error
-import urllib.parse
+import urllib3
 import time
 
 import cli
 
 from cli.exceptions import CLIException
 
+urllib3.disable_warnings()
 
-def read_endpoint(addr, endpoint, query=None):
+def read_endpoint(addr, endpoint, config, query=None):
     """
     Read the specified endpoint and return the results.
     """
@@ -38,20 +37,23 @@ def read_endpoint(addr, endpoint, query=None):
     except Exception as exception:
         raise CLIException("Unable to sanitize address '{addr}': {error}"
                            .format(addr=addr, error=str(exception)))
-
     try:
         url = "{addr}/{endpoint}".format(addr=addr, endpoint=endpoint)
         if query is not None:
             url += "?{query}".format(query=urllib.parse.urlencode(query))
-        http_response = urllib.request.urlopen(url).read().decode("utf-8")
+        headers = urllib3.make_headers(basic_auth=config.principal() + ":" + config.secret())
+        http = urllib3.PoolManager()
+        http_response = http.request('GET', url, headers=headers)
     except Exception as exception:
+        print(exception)
         raise CLIException("Unable to open url '{url}': {error}"
                            .format(url=url, error=str(exception)))
 
-    return http_response
+
+    return http_response.data.decode('utf-8')
 
 
-def get_json(addr, endpoint, condition=None, timeout=5, query=None):
+def get_json(addr, endpoint, config, condition=None, timeout=5, query=None):
     """
     Return the contents of the 'endpoint' at 'addr' as JSON data
     subject to the condition specified in 'condition'. If we are
@@ -64,7 +66,7 @@ def get_json(addr, endpoint, condition=None, timeout=5, query=None):
         data = None
 
         try:
-            data = read_endpoint(addr, endpoint, query)
+            data = read_endpoint(addr, endpoint, config, query)
         except Exception as exception:
             pass
 
--- a/src/python/cli_new/lib/cli/http.py
+++ b/src/python/cli_new/lib/cli/http.py
@@ -19,19 +19,22 @@ A collection of http related functions used by the CLI and its Plugins.
 """
 
 import json
+from urllib.parse import urlencode
 import urllib3
-import time
 
 import cli
 
 from cli.exceptions import CLIException
 
+# Disable all SSL warnings. These are not necessary, as the user has
+# the option to disable SSL verification.
 urllib3.disable_warnings()
 
 def read_endpoint(addr, endpoint, config, query=None):
     """
     Read the specified endpoint and return the results.
     """
+
     try:
         addr = cli.util.sanitize_address(addr)
     except Exception as exception:
@@ -40,51 +43,46 @@ def read_endpoint(addr, endpoint, config, query=None):
     try:
         url = "{addr}/{endpoint}".format(addr=addr, endpoint=endpoint)
         if query is not None:
-            url += "?{query}".format(query=urllib.parse.urlencode(query))
-        headers = urllib3.make_headers(basic_auth=config.principal() + ":" + config.secret())
+            url += "?{query}".format(query=urlencode(query))
+        if config.principal() is not None and config.secret() is not None:
+            headers = urllib3.make_headers(
+                basic_auth=config.principal() + ":" + config.secret()
+            )
+        else:
+            headers = None
         http = urllib3.PoolManager()
-        http_response = http.request('GET', url, headers=headers)
+        http_response = http.request(
+            'GET',
+            url,
+            headers=headers,
+            timeout=config.agent_timeout()
+        )
+        return http_response.data.decode('utf-8')
+
     except Exception as exception:
-        print(exception)
         raise CLIException("Unable to open url '{url}': {error}"
                            .format(url=url, error=str(exception)))
 
 
-    return http_response.data.decode('utf-8')
-
-
-def get_json(addr, endpoint, config, condition=None, timeout=5, query=None):
+def get_json(addr, endpoint, config, condition=None, query=None):
     """
     Return the contents of the 'endpoint' at 'addr' as JSON data
     subject to the condition specified in 'condition'. If we are
-    unable to read the data or unable to meet the condition within
-    'timeout' seconds we throw an error.
+    unable to read the data we throw an error.
     """
-    start_time = time.time()
 
-    while True:
-        data = None
+    data = read_endpoint(addr, endpoint, config, query)
 
-        try:
-            data = read_endpoint(addr, endpoint, config, query)
-        except Exception as exception:
-            pass
+    try:
+        data = json.loads(data)
+    except Exception as exception:
+        raise CLIException("Could not load JSON from '{data}': {error}"
+                           .format(data=data, error=str(exception)))
 
-        if data:
-            try:
-                data = json.loads(data)
-            except Exception as exception:
-                raise CLIException("Could not load JSON from '{data}': {error}"
-                                   .format(data=data, error=str(exception)))
+    if not condition:
+        return data
 
-            if not condition:
-                return data
+    if condition(data):
+        return data
 
-            if condition(data):
-                return data
-
-        if time.time() - start_time > timeout:
-            raise CLIException("Failed to get data within {seconds} seconds"
-                               .format(seconds=str(timeout)))
-
-        time.sleep(0.1)
+    return data
--- a/src/python/cli_new/lib/cli/mesos.py
+++ b/src/python/cli_new/lib/cli/mesos.py
@@ -120,6 +120,9 @@ def get_tasks(master, config, query=None):
     endpoint = "tasks"
     key = "tasks"
 
+    if query is None:
+        query = {'order':'asc'}
+
     try:
         data = http.get_json(master, endpoint, config, query=query)
     except Exception as exception:
@@ -161,7 +164,7 @@ class TaskIO():
     HEARTBEAT_INTERVAL = 30
     HEARTBEAT_INTERVAL_NANOSECONDS = HEARTBEAT_INTERVAL * 1000000000
 
-    def __init__(self, master, task_id):
+    def __init__(self, master, config, task_id):
         # Get the task and make sure its container was launched by the UCR.
         # Since task's containers are launched by the UCR by default, we want
         # to allow most tasks to pass through unchecked. The only exception is
@@ -169,7 +172,7 @@ class TaskIO():
         # "MESOS". Having a type of "MESOS" implies that it was launched by the
         # UCR -- all other types imply it was not.
         try:
-            tasks = get_tasks(master, query={'task_id': task_id})
+            tasks = get_tasks(master, config, query={'task_id': task_id})
         except Exception as exception:
             raise CLIException("Unable to get task with ID {task_id}"
                                " from leading master '{master}': {error}"
@@ -198,11 +201,13 @@ class TaskIO():
                         "This command is only supported for tasks"
                         " launched by the Universal Container Runtime (UCR).")
 
+        # Get the scheme of the agent
+        scheme = "https://" if config.agent_ssl() else "http://"
+
         # Get the URL to the agent running the task.
         agent_addr = util.sanitize_address(
-            get_agent_address(task_obj["slave_id"], master))
+            scheme + get_agent_address(task_obj["slave_id"], master, config))
         self.agent_url = mesos.http.simple_urljoin(agent_addr, "api/v1")
-
         # Get the agent's task path by checking the `state` endpoint.
         try:
             self.container_id = get_container_id(task_obj)
@@ -252,6 +257,7 @@ class TaskIO():
         self.interactive = False
         self.tty = False
         self.output_thread_entry_point = None
+        self.config = config
 
         # Allow an exit sequence to be used to break the CLIs attachment to
         # the remote task. Depending on the call, this may be disabled, or
@@ -322,7 +328,6 @@ class TaskIO():
         self.args = _args
         self.interactive = _interactive
         self.tty = _tty
-
         # Override the container ID with the current container ID as the
         # parent, and generate a new UUID for the nested container used to
         # run commands passed to `task exec`.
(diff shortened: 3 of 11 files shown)


src/python/cli_new/lib/cli/http.py
```diff
@@ -64,7 +73,7 @@ def get_json(addr, endpoint, condition=None, timeout=5, query=None):
         data = None
 
         try:
-            data = read_endpoint(addr, endpoint, query)
+            data = read_endpoint(addr, endpoint, config, query)
         except Exception as exception:
```
asekretenko, line 77: Well, I understand that this code used to silently drop arbitrary errors since time immemorial, but probably now it is the right time to fix this:)

src/python/cli_new/lib/cli/mesos.py
```diff
@@ -504,13 +508,23 @@ def _attach_container_output(self):
...
 
+        # Set authentication header
+        auth = None
+        # pylint: disable=line-too-long
+        if self.config.agent_principal() is not None and self.config.agent_secret() is not None:
+            auth = requests.auth.HTTPBasicAuth(
+                self.config.agent_principal(),
+                self.config.agent_secret()
+            )
+
```
asekretenko, line 519: Given that this code occurs three times, it is definitely worth extracting into a dedicated function.

src/python/cli_new/lib/cli/mesos.py
```diff
@@ -539,6 +554,16 @@ def _launch_nested_container_session(self):
         nested container and attach to its output stream.
         The output stream is then sent back in the response.
         """
+
+        # Set authentication header
+        # pylint: disable=line-too-long
+        auth = None
+        if self.config.agent_principal() is not None and self.config.agent_secret() is not None:
```
asekretenko, line 561: You can just wrap the line instead of suppressing pylint warnings:
```
if self.config.agent_principal() is not None and \
    self.config.agent_secret() is not None:
```

asekretenko (commented):
The approach looks good overall; thanks for addressing this!

src/python/cli_new/lib/cli/config.py
```diff
@@ -119,6 +119,65 @@ def master(self):
 
         return master
 
+    def principal(self):
+        """
+        Return the principal in the configuration file
+        """
+        if "principal" not in self.data["master"]:
+            return None
```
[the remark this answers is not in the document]
andreaspeters: I add the check if the parameter set and not empty. Do u have other checks in your mind? :-)

src/python/cli_new/lib/cli/mesos.py
andreaspeters: U are right. I also moved it into the config.py. but I'm unsteady, maybe it makes more sense in http.py. :thinking:

src/python/cli_new/lib/cli/http.py
```diff
@@ -38,20 +38,29 @@ def read_endpoint(addr, endpoint, query=None):
...
+        if config.principal() is not None and config.secret() is not None:
+            headers = urllib3.make_headers(
+                basic_auth=config.principal() + ":" + config.secret()
+            )
+        else:
+            headers = None
+        http = urllib3.PoolManager()
+        http_response = http.request('GET', url, headers=headers)
     except Exception as exception:
+        print(exception)
```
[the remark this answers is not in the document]
andreaspeters: right, I move it into the try->exception! :-)

src/python/cli_new/lib/cli/http.py
andreaspeters: Well, good that you speak it out. The loop also bother me. So I removed it and add a timeout into the read_endpoint function. :-) Hope it looks good now.

andreaspeters:
Does we have a list of features we want to see in the mesos_cli? I'm also planning to develope a plugin for our kubernetes framework.

andreaspeters:
@asekretenko I'm done now. U can review again. :-)

src/python/cli_new/lib/cli/config.py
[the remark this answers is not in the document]
cf-natali: Expanding on @asekretenko suggestion, this might be better:
```
principal = self.data["master"].get("principal")
if principal is None:
    return None
elif not principal:
    raise CLIException("The 'principal' field in 'master'"
                                   " must be non-empty")
return principal
```

Same thing for the other accessors.

src/python/cli_new/lib/cli/http.py
```diff
@@ -19,70 +19,71 @@
...
 
+urllib3.disable_warnings()
 
-def read_endpoint(addr, endpoint, query=None):
+def read_endpoint(addr, endpoint, config, query=None):
     """
     Read the specified endpoint and return the results.
     """
+
+    data = None
```
cf-natali, line 36: Looking at the code below it doesn't seem that the function can ever return `None` for `data` - maybe no need to set it to `None` here, and just directly return below:
```
return http_response.data.decode('utf-8')
```

src/python/cli_new/lib/cli/http.py
```diff
@@ -19,70 +19,71 @@
...
-        if data:
-            try:
-                data = json.loads(data)
-            except Exception as exception:
-                raise CLIException("Could not load JSON from '{data}': {error}"
-                                   .format(data=data, error=str(exception)))
+    try:
+        data = read_endpoint(addr, endpoint, config, query)
+    except Exception as exception:
+        pass
```
cf-natali, line 75: Hm, I'm not sure I understand what's going on here - the function's docstring says "If we are unable to read the data we throw an error.", so why catch the exception?
And also, if we catch an exception here, `data` will be `None`, so the below `data = json.loads(data)` is definitely not going to work.

src/python/cli_new/lib/cli/mesos.py
```diff
@@ -114,15 +113,16 @@ def get_container_id(task):
         " Please try again.")
 
 
-def get_tasks(master, query=None):
+# pylint: disable=dangerous-default-value
```
cf-natali, line 116: Maybe instead of disabling the pylint warning which is legitimate it'd be better to default `query=None` and in the function's body do:
```
if query is None:
    query = {'order':'asc'}
```

src/python/cli_new/lib/cli/mesos.py
```diff
@@ -199,11 +199,14 @@ def __init__(self, master, task_id):
                         "This command is only supported for tasks"
                         " launched by the Universal Container Runtime (UCR).")
 
+        # Get the scheme of the agent
```
cf-natali, line 202: Might be more readable:
```
scheme = "https://" if config.agent_ssl() else "http://"
```

src/python/cli_new/lib/cli/http.py
andreaspeters: Sorry this part I just removed the loop around. I don't know what the idea behind this exception handling was. But I think u make a good point. Let me have a look to see if I can change.

src/python/cli_new/lib/cli/mesos.py
andreaspeters: done

src/python/cli_new/lib/cli/http.py
andreaspeters: yes, u are right. :+1:

src/python/cli_new/lib/cli/mesos.py
andreaspeters: Oh yes, that's pretty. Didn't know that. :-) Thanks @cf-natali

src/python/cli_new/lib/cli/http.py
andreaspeters: So, I just remove the exception handling. Like you said, I make so sense there. :+1:

kamaradclimber:
Could you squash all commits to increase readability of this history before merging?

andreaspeters:
> Could you squash all commits to increase readability of this history before merging?
Done :-)

kamaradclimber:
Are you sure? I still see 17 commits.

andreaspeters:
> Are you sure? I still see 17 commits.

:thinking:  not 100%! But like u see, the last commit include all changes. :man_shrugging:

andreaspeters:
> > Are you sure? I still see 17 commits.
> 
> thinking not 100%! But like u see, the last commit include all changes. man_shrugging

Yeah it looks like I done something wrong during my past commits that the squash didn't worked as I expected. Can u squash during the MR?

cf-natali:
> > > Are you sure? I still see 17 commits.
> > 
> > 
> > thinking not 100%! But like u see, the last commit include all changes. man_shrugging
> 
> Yeah it looks like I done something wrong during my past commits that the squash didn't worked as I expected. Can u squash during the MR?

Maybe, but it'll be easier using `git rebase`. Then just need to `git push -f`

andreaspeters:
> > > > Are you sure? I still see 17 commits.
> > > 
> > > 
> > > thinking not 100%! But like u see, the last commit include all changes. man_shrugging
> > 
> > 
> > Yeah it looks like I done something wrong during my past commits that the squash didn't worked as I expected. Can u squash during the MR?
> 
> Maybe, but it'll be easier using `git rebase`. Then just need to `git push -f`

Hopefully I didn't broke more as I fixed with the rebase. :sweat_smile:

src/python/cli_new/lib/cli/config.py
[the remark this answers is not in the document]
cf-natali: Hm, you don't seem to have done this?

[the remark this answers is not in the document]
andreaspeters: I didn't feel it's a so big different to before. Thats why I ask! But I have no problem to change it. What do u think?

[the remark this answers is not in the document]
andreaspeters: Ok, I change it. :-) Shell I squash the commit again? :joy:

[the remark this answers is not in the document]
andreaspeters: Actually, I can also remove the Exception handling. The principal, and secret is not a "must-have".

src/python/cli_new/lib/cli/config.py
```diff
@@ -119,6 +120,94 @@ def master(self):
...
+    def agent_ssl(self):
+        """
+        Return if the agent support ssl
+        """
+        if "agent" in self.data:
+            agent_ssl = self.data["agent"].get("ssl")
+            if agent_ssl is None:
+                return None
+            elif not agent_ssl:
+                if not isinstance(self.data["agent"]["ssl"], bool):
```
cf-natali, line 158:
```
isinstance(agent_ssl, bool)
```

src/python/cli_new/lib/cli/http.py
```diff
@@ -19,70 +19,63 @@
...
-import urllib.parse
-import time
+from urllib.parse import urlencode
+import urllib3
 
 import cli
 
 from cli.exceptions import CLIException
 
+urllib3.disable_warnings()
```
cf-natali, line 29: Why is this necessary?

src/python/cli_new/lib/cli/http.py
```diff
@@ -19,70 +19,63 @@
...
-        http_response = urllib.request.urlopen(url).read().decode("utf-8")
+            url += "?{query}".format(query=urlencode(query))
+        if config.principal() is not None and config.secret() is not None:
+            headers = urllib3.make_headers(
+                basic_auth=config.principal() + ":" + config.secret()
+            )
+        else:
+            headers = None
+        http = urllib3.PoolManager()
+        http_response = http.request('GET', url, headers=headers, timeout=5)
```
cf-natali, line 52: Maybe make `timeout` a parameter with default 5 instead of hard-coding it, so it could be specified if needed.

src/python/cli_new/lib/cli/config.py
andreaspeters: Is already change in the current version.  But I'm interesting of what do u think about to change `def secret` and `def principal` to 

```
    def secret(self):
        return self.data["master"].get("secret")
```
No matter secret does not exist, or it's empty, secret would be "None". Thats why the "elif" is not usefull in these case. And because secret does not have to be configured, it's also not important to rise a "Exception".  What do u think?

src/python/cli_new/lib/cli/http.py
andreaspeters: I'm getting a lot of non pretty warning if I'm using a self signed certificate. But because of the possibility to set "ssl_verify", didn't think it made sense to issue the warning as well.

src/python/cli_new/lib/cli/http.py
andreaspeters: :+1: good idea. 
Even if it's not easy to see that someone "want" have so many changes in the code, I feel like the code is getting better and better. :-) The discuss about some points helping a lot. So, thank's @cf-natali @asekretenko @kamaradclimber

src/python/cli_new/lib/cli/config.py
cf-natali: > Is already change in the current version.

I don't seem to see it.

Also, I still see 3 commits, I thought you'd squashed?

cf-natali: > But I'm interesting of what do u think about to change `def secret` and `def principal` to
> 
> ```
>     def secret(self):
>         return self.data["master"].get("secret")
> ```
> 
> No matter secret does not exist, or it's empty, secret would be "None". Thats why the "elif" is not usefull in these case. And because secret does not have to be configured, it's also not important to rise a "Exception". What do u think?

Yes, if it's not required it makes sense.

src/python/cli_new/lib/cli/http.py
cf-natali: Then maybe add a comment explaining it?

src/python/cli_new/lib/cli/config.py
andreaspeters: :+1:

src/python/cli_new/lib/cli/http.py
andreaspeters: Will do it :-)

src/python/cli_new/lib/cli/config.py
andreaspeters: Yeah I didnt commit and pushed just now. I wanted to wait until I get the feedback to the "def secret". :-)  Then I will also squash again.

src/python/cli_new/lib/cli/config.py
```diff
@@ -119,6 +120,92 @@ def master(self):
...
+        return None
+
+    def agent_timeout(self):
+        """
+        Return the connection timeout of the agent
+        """
+        if "agent" in self.data:
+            timeout = self.data["agent"].get("timeout")
+            if timeout is None:
+                return 5
```
cf-natali, line 174: I think you can do instead:
```
            timeout = self.data["agent"].get("timeout", 5)
```

Also, maybe better than hard-coding the default could be moved to the function prototype?
```
    def agent_timeout(self, default=5):
```

src/python/cli_new/lib/cli/config.py
```diff
@@ -119,6 +120,92 @@ def master(self):
...
+    def agent_principal(self):
+        """
+        Return the principal in the configuration file
+        """
+        if "agent" in self.data:
+            principal = self.data["agent"].get("principal")
+            if principal is None:
+                return None
+
+            return principal
```
cf-natali, line 192: This seems equivalent to just:
```
return self.data["agent"].get("principal")
```

src/python/cli_new/lib/cli/config.py
```diff
@@ -119,6 +120,92 @@ def master(self):
...
+        return None
+
+    def agent_secret(self):
+        """
+        Return the secret in the configuration file
+        """
+        if "agent" in self.data:
+            secret = self.data["agent"].get("secret")
+            if secret is None:
+                return None
```
cf-natali, line 203: This seems equivalent to just:
```
return self.data["agent"].get("secret")
```

src/python/cli_new/lib/cli/http.py
```diff
@@ -19,70 +19,70 @@
...
-            url += "?{query}".format(query=urllib.parse.urlencode(query))
-        http_response = urllib.request.urlopen(url).read().decode("utf-8")
+            url += "?{query}".format(query=urlencode(query))
+        if config.principal() is not None and config.secret() is not None:
+            headers = urllib3.make_headers(
+                basic_auth=config.principal() + ":" + config.secret()
+            )
+        else:
+            headers = None
+        http = urllib3.PoolManager()
```
cf-natali, line 53: I'm not familiar with `urllib3` - what does `Poolmanager` bring over just using  `utllib.request`?
Does it not support custom headers?

src/python/cli_new/lib/cli/mesos.py
```diff
@@ -413,6 +417,7 @@ def _wait(self):
             'wait_container': {
                 'container_id': self.container_id}}
         req_extra_args = {
+            'verify': self.config.agent_ssl_verify(),
```
cf-natali, line 420: What happens if `elf.config.agent_ssl_verify()` is `None`, will it work?

src/python/cli_new/lib/cli/plugins/task/main.py
```diff
@@ -75,11 +75,11 @@ def attach(self, argv):
         """
         try:
             master = self.config.master()
+            config = self.config
         except Exception as exception:
             raise CLIException("Unable to get leading master address: {error}"
                                .format(error=exception))
-
-        task_io = TaskIO(master, argv["<task-id>"])
+        task_io = TaskIO(master, config, argv["<task-id>"])
```
cf-natali, line 82: Nit: I think the indentation with the newline before was more readable.

src/python/cli_new/lib/cli/config.py
andreaspeters: :joy: I was not so brave. But of course, u are right.

src/python/cli_new/lib/cli/config.py
andreaspeters: I like the idea with the default value in the prototype. I will use it also for the agent_ssl and agent_ssl_verify. That will also simplify that functions. :+1:  thanks Charles.

src/python/cli_new/lib/cli/mesos.py
andreaspeters: Thats obsolete now :-)

ridv:
> Could you squash all commits to increase readability of this history before merging?

In the past the commiter merging the change used the "squash and merge" strategy which has the same effect. So no need to do anything special  😄 . (In fact, it's probably better to leave the individual commits to see how the patch progressed which is a bit of a carry over from when reviews happened on Review Board)

src/python/cli_new/lib/cli/config.py
```diff
@@ -119,6 +120,79 @@ def master(self):
...
+        """
+        if "agent" in self.data:
+            agent_ssl = self.data["agent"].get("ssl", default)
+            if not isinstance(agent_ssl, bool):
+                raise CLIException("The 'agent->ssl' field"
+                                   " must be True/False")
+
+            return agent_ssl
+
+        return default
```
cf-natali, line 147: So you can simplify it even further by using a default for `self.data["agent"], like this]:
```
def agent_ssl(self, default=False):
    agent_ssl = self.data.get("agent", {}).get("ssl", default)
    if not isinstance(agent_ssl, bool):
        raise CLIException("The 'agent->ssl' field must be True/False"
    return agent_ssl
```

Personally I like it because it removes a repetition between checking that `"agent"` is in `data` and then accessing it, and also it removes a branch, but up to you if you find it less natural.

src/python/cli_new/lib/cli/config.py
```diff
@@ -119,6 +120,79 @@ def master(self):
...
+
+    def agent_ssl_verify(self, default=False):
+        """
+        Return if the ssl certificate should be verified
+        """
+        if "agent" in self.data:
```
cf-natali, line 153: Same here.

src/python/cli_new/lib/cli/config.py
```diff
@@ -137,3 +211,15 @@ def plugins(self):
             return self.data["plugins"]
 
         return []
+
+    def authentication_header(self):
+        """
+        Return the BasicAuth authentication header
+        """
+        if (self.agent_principal() is not None
+                and self.agent_secret() is not None):
```
cf-natali, line 220: Nit: the indentation seems a bit strange?

src/python/cli_new/lib/cli/plugins/task/main.py
```diff
@@ -106,12 +108,13 @@ def list(self, argv):
         # pylint: disable=unused-argument
         try:
             master = self.config.master()
+            config = self.config
```
cf-natali, line 111: Why store it here and not directly pass `self.config` below?
```
        try:
            tasks = get_tasks(master, config)
```

src/python/cli_new/lib/cli/config.py
andreaspeters: I also was thinking to replace it like that, but in my eyes it looks too complicated. :thinking: You don't think so?

src/python/cli_new/lib/cli/config.py
andreaspeters: The linter say it's fine like that. :man_shrugging:  And my vim say, it should be like that. :joy:

src/python/cli_new/lib/cli/plugins/task/main.py
andreaspeters: I thinks it's easier to read.

cf-natali: OK!

src/python/cli_new/lib/cli/config.py
cf-natali: Up to you, I'm fine either way :).

cf-natali (approved):

cf-natali:
@asekretenko @qianzhangxa 
Maybe you could have a look at it?

andreaspeters:
> @asekretenko @qianzhangxa
> Maybe you could have a look at it?

Some feedback? :-)

qianzhangxa:
I am not very familiar with Python :(
@asekretenko can you please review again?

src/python/cli_new/lib/cli/http.py
```diff
@@ -19,70 +19,70 @@
...
 
-            if condition(data):
-                return data
+    if not condition:
+        return data
 
-        if time.time() - start_time > timeout:
-            raise CLIException("Failed to get data within {seconds} seconds"
-                               .format(seconds=str(timeout)))
+    if condition(data):
```
asekretenko, line 85: Does this mean that this code now returns `data` anyway, regardless of whether `condition(data)` returned `true` or `false`? Is the `condition` parameter really needed then?

It occurs to me that we have a lot of code in the tests which used to use this function to wait until `condition` becomes true (example: https://github.com/apache/mesos/blob/72f16f68973bf7d2ce5c621539a21fc4eccfa56e/src/python/cli_new/lib/cli/tests/base.py#L280).

While removing a lame retry loop is a good thing (this loop in its existing form is harmful for production environments, as it contains no exponential backoff logic), something needs to be done with these tests make use of `condition`.

If these tests are dead code, that dead code should be removed (ideally, in a separate commit). 
On the other hand, if this wait-for-condition functionality is really needed by tests, I would say that it should be pulled into a separate function wrapping the new `get_json()`. Given that this is apparently only used in tests, this new wrapper fucnction should be a part of the tests suite and not of this file.

asekretenko:
Other than this question about the `condition` logic, I don't see any other issues.

src/python/cli_new/lib/cli/http.py
andreaspeters: Actually I'm not 100% sure what the idea behind `condition` was. But yes, `data` have ever be to returned. But if `condition` is set, then through that function. I will create (for myself) a issue to investigate `condition` to see if it's needed or if we can remove it. Then we can merge this PR. Is this procedure ok for everyone? :-)

andreaspeters: Actually I'm not 100% sure what the idea behind condition was. But yes, data have ever be to returned. But if condition is set, then through that function. I will create (for myself) a issue to investigate condition to see if it's needed or if we can remove it. Then we can merge this PR. Is this procedure ok for everyone? :-)

andreaspeters:
Can we merge it please?

qianzhangxa:
Sorry for the late response, I have merged it, thanks for the contributions @andreaspeters!

andreaspeters:
No problem @qianzhangxa. Your welcome. :-)

PR #385, opened 2021-05-15
Fixed read of uninitialized variables detected with UBSAN.

cf-natali:
@asekretenko @qianzhangxa 

They were detected as invalid boolean values.

I'll look at adding a `configure` flag to build with UBSAN, I need to look at potential remaining false positives.

### DIFF
--- a/src/tests/cluster.hpp
+++ b/src/tests/cluster.hpp
@@ -106,7 +106,8 @@ public:
   void setAuthorizationCallbacks(Authorizer* authorizer);
 
 private:
-  Master() : files(master::READONLY_HTTP_AUTHENTICATION_REALM) {};
+  Master() : files(master::READONLY_HTTP_AUTHENTICATION_REALM),
+             authorizationCallbacksSet(false) {}
 
   // Not copyable, not assignable.
   Master(const Master&) = delete;
@@ -203,7 +204,8 @@ public:
   void setAuthorizationCallbacks(Authorizer* authorizer);
 
 private:
-  Slave() : files(slave::READONLY_HTTP_AUTHENTICATION_REALM) {};
+  Slave() : files(slave::READONLY_HTTP_AUTHENTICATION_REALM),
+            authorizationCallbacksSet(false) {}
 
   // Not copyable, not assignable.
   Slave(const Slave&) = delete;


asekretenko (approved):

cf-natali:
@asekretenko Would it be possible to revert it?
Since merging it I've noticed that I get some tests failing randomly - it's really strange, they are cgroup isolator tests which seem completely unrelated, e.g.:

```
[ RUN      ] MesosContainerizerTest.StatusWithContainerID
../../src/tests/mesos.cpp:736: Failure
Failed to wait 15secs for cgroups::destroy(hierarchy, cgroup)
-----------------------------------------------------------
We're very sorry but we can't seem to destroy existing
cgroups that we likely created as part of an earlier
invocation of the tests. Please manually destroy the cgroup
at '/sys/fs/cgroup/memory/mesos_test_92a75d38-a749-4894-9746-79f8b3590b7c/57721582-cd4d-43a7-8fcf-0b57e26a7987' by first
manually killing all the processes found in the file at '/sys/fs/cgroup/memory/mesos_test_92a75d38-a749-4894-9746-79f8b3590b7c/57721582-cd4d-43a7-8fcf-0b57e26a7987/tasks'
-----------------------------------------------------------
../../src/tests/mesos.cpp:782: Failure
Failed to wait 15secs for cgroups::destroy(hierarchy, cgroup)
[  FAILED  ] MesosContainerizerTest.StatusWithContainerID (30062 ms)
```

I have no idea what's going on, the only explanation I can see is that somehow before because these variables where not initialised they were pretty much always `true` (non-0 bool) and somehow the tests might depend on it.
I'll try to understand what causes it but in the meantime probably best to revert.

asekretenko:
@cf-natali I'm also observing this on some platforms in an internal CI; reverted until further investigation.

One more note: after looking at the code that makes use of those attributes more carefully, I'm realizing that our tests only work by miracle: the authorization callbacks that can be cleared both by `~Master()` and by `~Slave()` belong to global libprocess state. Given that we definitely have tests that are destroying slaves and masters in the middle of the test, I'm now suprrised how this stuff works at all... most likely, we just don't have tests that both rely on authn callbacks and destroy a master/an agent in the middle. Not sure if this story has any relation to the cgroup destruction failure, though.

cf-natali:
Oh damn, I had a look and it looks like the cgroup is a separate problem, which looks like a kernel bug...
I'll dig.

cf-natali:
Found the cause of the cgroup issue: https://github.com/apache/mesos/pull/388
Kernel update triggering a bug in the cgroup tasks killing.

PR #387, opened 2021-05-16
Fixed compilation error on CAP_(PERFMON,BPF,CHECKPOINT_RESTORE).

cf-natali:
They should be defined without the "CAP_" prefix to avoid clashing with
the corresponding definitions in `linux/capability.h` - and be
consistent with other capabilities.

@asekretenko

### DIFF
--- a/include/mesos/mesos.proto
+++ b/include/mesos/mesos.proto
@@ -3378,9 +3378,9 @@ message CapabilityInfo {
     WAKE_ALARM = 1035;
     BLOCK_SUSPEND = 1036;
     AUDIT_READ = 1037;
-    CAP_PERFMON = 1038;
-    CAP_BPF = 1039;
-    CAP_CHECKPOINT_RESTORE = 1040;
+    PERFMON = 1038;
+    BPF = 1039;
+    CHECKPOINT_RESTORE = 1040;
   }
 
   repeated Capability capabilities = 1;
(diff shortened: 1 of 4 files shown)


andreaspeters (approved):

asekretenko:
@cf-natali It would help readers if the commit message made it clear that the issue is only triggered by recent linux headers (something in line of "Fixed compilation against linux 5.0+ capabilities.h" - I'm not sure which exact introduced them, please check that). Otherwise, looks good to me. 

Great that you caught this before releasing this code! Otherwise, this rename in `proto`s would be potentially breaking users of json APIs.

cf-natali:
> @cf-natali It would help readers if the commit message made it clear that the issue is only triggered by recent linux headers (something in line of "Fixed compilation against linux 5.0+ capabilities.h" - I'm not sure which exact kernel version introduced them, please check that). Otherwise, looks good to me.

Done!

cf-natali:
> Great that you caught this before releasing this code! Otherwise, this rename in protos would be potentially breaking users of json APIs (due to how our json deserialization uses protobuf reflection).

Yeah, although it was my fault in the first place, at the time I didn't have access to such a recent kernel/libc, so couldn't properly test it.
It'd help if we had more recent/varied buildbots.

cf-natali:
@qianzhangxa

asekretenko (approved):

PR #388, opened 2021-05-17
Fixed a bug where the cgroup task killer leaves the cgroup frozen.

cf-natali:
When its future is discarded, the cgroup task killer would stop immediately. This potentially leaves the cgroup frozen, therefore leaving all processes in uninterruptible state, which is quite bad - for example the tasks would be unkillable and the cgroup couldn't be destroyed.
Instead, when discarded, wait a bit to give the task killer a chance to finish cleanly, killing all the tasks and thawing the cgroup.

I started seeing this after updating my kernel - dozens of tests would fail randomly when trying to destroy cgroups, e.g.:
```
[ RUN      ] SlaveRecoveryTest/0.ExecutorDanglingLatestSymlink
I0517 22:39:10.577574 573547 exec.cpp:164] Version: 1.12.0
I0517 22:39:10.591667 573550 exec.cpp:237] Executor registered on agent a7c3a499-2968-4eb6-91dd-1020fedc1522-S0
I0517 22:39:10.594743 573552 executor.cpp:190] Received SUBSCRIBED event
I0517 22:39:10.595999 573552 executor.cpp:194] Subscribed executor on thinkpad
I0517 22:39:10.596259 573552 executor.cpp:190] Received LAUNCH event
I0517 22:39:10.597656 573552 executor.cpp:722] Starting task 39b1539d-9c30-41b0-b643-3ece3e63bea5
I0517 22:39:10.617466 573552 executor.cpp:740] Forked command at 573554
../../src/tests/mesos.cpp:782: Failure
Failed to wait 15secs for cgroups::destroy(hierarchy, cgroup)
*** Aborted at 1621287565 (unix time) try "date -d @1621287565" if you are using GNU date ***
PC: @     0x56129f5437e7 testing::UnitTest::AddTestPartResult()
*** SIGSEGV (@0x0) received by PID 573473 (TID 0x7fb79b027b00) from PID 0; stack trace: ***
    @     0x7fb79c287140 (unknown)
    @     0x56129f5437e7 testing::UnitTest::AddTestPartResult()
    @     0x56129f5366f3 testing::internal::AssertHelper::operator=()
    @     0x56129e60421b mesos::internal::tests::ContainerizerTest<>::TearDown()
    @     0x56129f5633cb testing::internal::HandleSehExceptionsInMethodIfSupported<>()
    @     0x56129f55d634 testing::internal::HandleExceptionsInMethodIfSupported<>()
    @     0x56129f53d38d testing::Test::Run()
    @     0x56129f53dbb2 testing::TestInfo::Run()
    @     0x56129f53e1f9 testing::TestCase::Run()
    @     0x56129f544ce0 testing::internal::UnitTestImpl::RunAllTests()
    @     0x56129f564349 testing::internal::HandleSehExceptionsInMethodIfSupported<>()
    @     0x56129f55e24a testing::internal::HandleExceptionsInMethodIfSupported<>()
    @     0x56129f543a0a testing::UnitTest::Run()
    @     0x56129e0f6aba RUN_ALL_TESTS()
    @     0x56129e0f6505 main
    @     0x7fb79c0d4d0a __libc_start_main
    @     0x56129d23545a _start
Erreur de segmentation (core dumped)
```

Comparing a successful run:
```
I0516 23:58:25.918263 466660 cgroups.cpp:2934] Freezing cgroup /sys/fs/cgroup/freezer/mesos_test_8851f242-b14a-4580-88b8-5d3dd552bfde/e2bf97c3-14a1-4b2b-b399-12dbd0f1c99d
I0516 23:58:25.918447 466657 cgroups.cpp:1323] Successfully froze cgroup /sys/fs/cgroup/freezer/mesos_test_8851f242-b14a-4580-88b8-5d3dd552bfde/e2bf97c3-14a1-4b2b-b399-12dbd0f1c99d after 157952ns
I0516 23:58:25.918900 466660 cgroups.cpp:2952] Thawing cgroup /sys/fs/cgroup/freezer/mesos_test_8851f242-b14a-4580-88b8-5d3dd552bfde/e2bf97c3-14a1-4b2b-b399-12dbd0f1c99d
I0516 23:58:25.919064 466656 cgroups.cpp:1352] Successfully thawed cgroup /sys/fs/cgroup/freezer/mesos_test_8851f242-b14a-4580-88b8-5d3dd552bfde/e2bf97c3-14a1-4b2b-b399-12dbd0f1c99d after 134912ns
```

To an unsuccessful one:
```
I0516 23:58:42.638830 466897 linux_launcher.cpp:606] Destroying cgroup '/sys/fs/cgroup/freezer/mesos_test_98d2814b-03ad-4c14-a3f9-832ab688e386/3139257f-4e7a-4828-9af4-75c867846558'
I0516 23:58:42.639083 466898 composing.cpp:343] Finished recovering all containerizers
I0516 23:58:42.639549 466898 cgroups.cpp:2934] Freezing cgroup /sys/fs/cgroup/freezer/mesos_test_98d2814b-03ad-4c14-a3f9-832ab688e386/3139257f-4e7a-4828-9af4-75c867846558
I0516 23:58:42.714794 466899 cgroups.cpp:2952] Thawing cgroup /sys/fs/cgroup/freezer/mesos_test_98d2814b-03ad-4c14-a3f9-832ab688e386/3139257f-4e7a-4828-9af4-75c867846558
W0516 23:58:42.745649 466903 cgroups.cpp:294] Removal of cgroup /sys/fs/cgroup/memory/mesos_test_98d2814b-03ad-4c14-a3f9-832ab688e386/3139257f-4e7a-4828-9af4-75c867846558 failed with EBUSY, will try again
W0516 23:58:42.746997 466902 cgroups.cpp:294] Removal of cgroup /sys/fs/cgroup/memory/mesos_test_98d2814b-03ad-4c14-a3f9-832ab688e386/3139257f-4e7a-4828-9af4-75c867846558 failed with EBUSY, will try again
W0516 23:58:42.749405 466898 cgroups.cpp:294] Removal of cgroup /sys/fs/cgroup/memory/mesos_test_98d2814b-03ad-4c14-a3f9-832ab688e386/3139257f-4e7a-4828-9af4-75c867846558 failed with EBUSY, will try again
W0516 23:58:42.754009 466900 cgroups.cpp:294] Removal of cgroup /sys/fs/cgroup/memory/mesos_test_98d2814b-03ad-4c14-a3f9-832ab688e386/3139257f-4e7a-4828-9af4-75c867846558 failed with EBUSY, will try again
W0516 23:58:42.762802 466901 cgroups.cpp:294] Removal of cgroup /sys/fs/cgroup/memory/mesos_test_98d2814b-03ad-4c14-a3f9-832ab688e386/3139257f-4e7a-4828-9af4-75c867846558 failed with EBUSY, will try again
W0516 23:58:42.779531 466904 cgroups.cpp:294] Removal of cgroup /sys/fs/cgroup/memory/mesos_test_98d2814b-03ad-4c14-a3f9-832ab688e386/3139257f-4e7a-4828-9af4-75c867846558 failed with EBUSY, will try again
W0516 23:58:42.812022 466897 cgroups.cpp:294] Removal of cgroup /sys/fs/cgroup/memory/mesos_test_98d2814b-03ad-4c14-a3f9-832ab688e386/3139257f-4e7a-4828-9af4-75c867846558 failed with EBUSY, will try again
W0516 23:58:42.877116 466903 cgroups.cpp:294] Removal of cgroup /sys/fs/cgroup/memory/mesos_test_98d2814b-03ad-4c14-a3f9-832ab688e386/3139257f-4e7a-4828-9af4-75c867846558 failed with EBUSY, will try again
W0516 23:58:43.006693 466898 cgroups.cpp:294] Removal of cgroup /sys/fs/cgroup/memory/mesos_test_98d2814b-03ad-4c14-a3f9-832ab688e386/3139257f-4e7a-4828-9af4-75c867846558 failed with EBUSY, will try again
W0516 23:58:43.264366 466899 cgroups.cpp:294] Removal of cgroup /sys/fs/cgroup/memory/mesos_test_98d2814b-03ad-4c14-a3f9-832ab688e386/3139257f-4e7a-4828-9af4-75c867846558 failed with EBUSY, will try again
W0516 23:58:43.777858 466901 cgroups.cpp:294] Removal of cgroup /sys/fs/cgroup/memory/mesos_test_98d2814b-03ad-4c14-a3f9-832ab688e386/3139257f-4e7a-4828-9af4-75c867846558 failed with EBUSY, will try again
W0516 23:58:44.803258 466897 cgroups.cpp:294] Removal of cgroup /sys/fs/cgroup/memory/mesos_test_98d2814b-03ad-4c14-a3f9-832ab688e386/3139257f-4e7a-4828-9af4-75c867846558 failed with EBUSY, will try again
W0516 23:58:46.851841 466901 cgroups.cpp:294] Removal of cgroup /sys/fs/cgroup/memory/mesos_test_98d2814b-03ad-4c14-a3f9-832ab688e386/3139257f-4e7a-4828-9af4-75c867846558 failed with EBUSY, will try again
```

We can see that when the problem occur, it's because the cgroup doesn't freeze quickly. Since the tests quickly destroys the slave/containeriser after the task finished, the cgroups task killer gets interrupted after the cgroup has started to be frozen but before it's been thawed (after killing the tasks) - https://github.com/apache/mesos/blob/master/src/linux/cgroups.cpp#L1445

```
  void killTasks() {
    // Chain together the steps needed to kill all tasks in the cgroup.
    chain = freeze()                     // Freeze the cgroup.
      .then(defer(self(), &Self::kill))  // Send kill signal.
      .then(defer(self(), &Self::thaw))  // Thaw cgroup to deliver signal.
      .then(defer(self(), &Self::reap)); // Wait until all pids are reaped.

    chain.onAny(defer(self(), &Self::finished, lambda::_1));
  }
```

The problem is that the process sets up a discard callback which immediately terminates the process - https://github.com/apache/mesos/blob/master/src/linux/cgroups.cpp#L1407
```
  void initialize() override
  {
    // Stop when no one cares.
    promise.future().onDiscard(lambda::bind(
        static_cast<void (*)(const UPID&, bool)>(terminate), self(), true));

    killTasks();
  }
```

Which means that the task killer can be interrupted after it's frozen the cgroup, but before killing the tasks and thawing.
Which means that the cgroup stays frozen, tasks are stuck in uninterruptible state, can't be killed and the cgroup can't be destroyed, as can be seen above.

I'm not adding a specific test because it's causing dozens of tests to fail already.

I'm not quite sure why this only started happening on recent kernels, but my guess is that for whatever reason freezing seems to sometime take longer than before, which means we're much more likely to interrupt the killer in the middle of its work.
In any case it's a long-standing bug which has potentially bad consequences so good to fix.

@asekretenko 
@qianzhangxa

### DIFF
--- a/src/linux/cgroups.cpp
+++ b/src/linux/cgroups.cpp
@@ -1398,15 +1398,14 @@ public:
   // Return a future indicating the state of the killer.
   // Failure occurs if any process in the cgroup is unable to be
   // killed.
+  // Discarding the future will cause this process to stop the next time it
+  // calls `freeze`: we don't want to stop at an arbitrary point since it might
+  // leave the cgroup frozen.
   Future<Nothing> future() { return promise.future(); }
 
 protected:
   void initialize() override
   {
-    // Stop when no one cares.
-    promise.future().onDiscard(lambda::bind(
-        static_cast<void (*)(const UPID&, bool)>(terminate), self(), true));
-
     killTasks();
   }
 
@@ -1425,16 +1424,15 @@ private:
       const PID<TasksKiller>& pid)
   {
     // Cancel the freeze operation.
-    // TODO(jieyu): Wait until 'future' is in DISCARDED state before
-    // starting retry.
     future.discard();
 
-    // We attempt to kill the processes before we thaw again,
-    // due to a bug in the kernel. See MESOS-1758 for more details.
-    // We thaw the cgroup before trying to freeze again to allow any
-    // pending signals to be delivered. See MESOS-1689 for details.
-    // This is a short term hack until we have PID namespace support.
-    return Future<bool>(true)
+    // Wait until the freeze is cancelled, and then attempt to kill the
+    // processes before we thaw again, due to a bug in the kernel. See
+    // MESOS-1758 for more details.  We thaw the cgroup before trying to freeze
+    // again to allow any pending signals to be delivered. See MESOS-1689 for
+    // details.  This is a short term hack until we have PID namespace support.
+    return future
+      .recover([](const Future<Nothing>&){ return Future<Nothing>(Nothing()); })
       .then(defer(pid, &Self::kill))
       .then(defer(pid, &Self::thaw))
       .then(defer(pid, &Self::freeze));
@@ -1452,6 +1450,12 @@ private:
 
   Future<Nothing> freeze()
   {
+    // Don't start another `killTasks` cycle if we've been asked to stop.
+    if (promise.future().hasDiscard()) {
+        terminate(self());
+        return Nothing();
+    }
+
     // TODO(jieyu): This is a workaround for MESOS-1689. We will move
     // away from freezer once we have pid namespace support.
     return cgroups::freezer::freeze(hierarchy, cgroup).after(


cf-natali:
Also maybe @abudnik if he's still willing to review cgroup code :).

src/linux/cgroups.cpp
```diff
@@ -1403,9 +1403,15 @@ class TasksKiller : public Process<TasksKiller>
...
-    promise.future().onDiscard(lambda::bind(
-        static_cast<void (*)(const UPID&, bool)>(terminate), self(), true));
+    // We don't want to stop immediately upon discard, because
+    // it could leave the cgroup frozen which means that processes
+    // are stuck in uninterrutible sleep (D state), which is quite bad.
+    // So upon discard we still do our best and keep trying to
+    // kill the cgroup for up to FREEZE_RETRY_INTERVAL which should be
+    // a reasonable upper bound.
+    promise.future().onDiscard([this]() {
+      delay(FREEZE_RETRY_INTERVAL, self(), &Self::selfTerminate);
```
kamaradclimber, line 1413: Would there be a way to avoid this extra delay if we detect the cgroup is not frozen at that time?

kamaradclimber (approved):

src/linux/cgroups.cpp
cf-natali: Hey @kamaradclimber , thanks for reviewing :).

It's just a delay for the discard: the killer process will continue running, which means that it will do its usual freeze/kill/thaw, and therefore will stop as soon as the tasks have been killed, it won't wait any longer than necessary. Since usually a cgroup can be frozen and killed very quickly - few ms - it should have negligible impact.

kamaradclimber: :ok_hand:

ridv:
This looks sane to me, and barring any opposition, this should probably be shipped. 9 days for such a small patch is already a lot. Can a Mesos committer either merge this or ask for improvements? This silence is exactly what I was afraid of when I voiced my concerns in the e-mail thread re: putting the project into the Attic.

cf-natali:
Yes.

Unless the involvement of existing committers significantly improves over
the next week or so, I'll contact the various users who manifested interest
in order to start a fork.

asekretenko:
@cf-natali I'm trying to look into this and to make sure this is not making some other issues worse.

For example, does this all mean that this deferred call that freezes a cgroup
https://github.com/apache/mesos/blob/85981d7c728798783b2e1e7cbed4f27a1497ccb2/src/linux/cgroups.cpp#L1440

has never been executed before this patch (due to immediate termination of a TaskKiller process), but, after this change, will be executing in some cases?...

(Btw, inablilty to comment on an untouched line in a github PR is one of examples why, as of 2019-2020, Mesos committers were still preferring to review contribution using Apache Reviewboard. I'm not saying that we should not do something to simplify contributing via GitHub, just illustrating why the RB has not been phased out.)

@ridv This, unfortunately, is not a "small patch", but an attempt to compensate for a deficiency in a workaround in a code that uses a rather problematic kernel interface and is already full of hacks and short-term solutions. I'm sorry, but reviewing such stuff meaningfully takes significant chunks of an uninterrupted time and should only be done in a sound mind and a good health (this was not my case in the beginnig of this week).

ridv:
@asekretenko that is absolutely understandable and I hope you find yourself in better health very soon. What you are doing here is greatly appreciated. 

My main concern, and this is in no way to put pressure on you as you're already doing a huge service to to project here, is the general radio silence from committers and PMC.

Either the community has enough reviewers and volunteers to keep the project going, or we don't and new people that want to continue the project need to be given permission to continue the project (as per the Apache board members during the Attic discussions). 

I think this is a conversation best had on the mailing list (and in fact, I've kicked off a thread).

cf-natali:
> @ridv This, unfortunately, is not a "small patch", but an attempt to compensate for a deficiency in a workaround in a code that uses a rather problematic kernel interface and is already full of hacks and short-term solutions.

Indeed, this freezer code is quite tricky, and with good reason - at work we experienced 3 different kernel bugs involving the freezer cgroup, it's really a mess even after all those years...

> @cf-natali I'm trying to look into this and to make sure this is not making some other issues worse.
> 
> For example, does this all mean that this deferred call that freezes a cgroup
> 
> https://github.com/apache/mesos/blob/85981d7c728798783b2e1e7cbed4f27a1497ccb2/src/linux/cgroups.cpp#L1440
> 
> has never been executed before this patch (due to immediate termination of a TaskKiller process), but, after this change, will be executing in some cases?...

As far as I can tell, this should not exercise any new code path, and should actually make unusual code paths less frequent.
The code you're referring to is only involved when the freeze operation timed out, and attempts to thaw and freeze it again - this can happen in normal circumstances, i.e. in the normal case where a cgroup is destroyed. This specific code path was exercised in the past - I checked using logs and `strace`, and will continue to be exercised if the freeze times out.

One way to view this change is that it can only delay the termination of the `TaskKiller` process: therefore, the `TaskKiller` will only terminate in a state in which it could already be terminated in in the current code.
The extra delay is a grace period letting some time to the `TaskKiller` to finish its freeze/kill/thaw cycle, hence making it much less likely to terminate it while it's in the middle of its freeze/kill/thaw.

> (Btw, inablilty to comment on an untouched line in a github PR is one of examples why, as of 2019-2020, Mesos committers were still preferring to review contribution using Apache Reviewboard. I'm not saying that we should not do something to simplify contributing via GitHub, just illustrating why the RB has not been phased out.)

Interesting, thanks!

> I'm sorry, but reviewing such stuff meaningfully takes significant chunks of an uninterrupted time and should only be done in a very sound mind and a very good health (this was not my case in the beginnig of this week).

I'm really sorry to hear that, and hope you're feeling better!

asekretenko:
@cf-natali After looking at this code more carefully and doing some experiments, I'm wondering if it is possible to more reliably prevent the root cause of this issue (leaving cgroup in a wrong state) from occurring.

If I'm not missing something, `TasksKiller::freeze()` essentially runs a retry loop, each iteration of which does the following:
1. calls `cgroups::freezer::freeze(hierarchy, cgroup)` limited by a timeout of a length determined by `FREEZE_RETRY_INTERVAL`
2. cancels timed-out result of the previous stage
3. calls `TasksKiller::kill()`
4. calls `TasksKiller::thaw()`, that is, `cgroups::freezer::thaw(hierarchy, cgroup)`
and then the loop repeats.
There are three conditions, upon which the loop ends:
a) success of `cgroups::freezer::freeze()`
b) failure at any stage
c) process termination caused by discard of TaskKiller's result future

That said, I'm not convinced that postponing c) is the best available option. 

Maybe we should consider replacing c) with another termination mechanism that will be **guaranteed** not to halt this loop in the middle of an iteration? Something like setting a flag in that discard callback and checking it before calling  `cgroups::freezer:freeze()` ?

Btw, how are you testing the code handling freeze/kill failures? 
In my experience, the simplest way to reliably create a process in an uninterruptable sleep (D state) is to create a file on some FUSE-backed FS (say, sshfs), mmap that file, stop the program backing the FS with SIGSTOP, and repeatedly read/write the mmap-ed bytes until the process eventually hangs. That's rather complicated, takes 30 seconds of read/write on my kernel and is not really usable in tests... I'm wondering if there is a simpler option.

cf-natali:
> That said, I'm not convinced that postponing c) is the best available option.
> 
> Maybe we should consider replacing c) with another termination mechanism that will be **guaranteed** not to halt this loop in the middle of an iteration? Something like ~setting a flag by that discard callback and checking this flag~ checking `future.hasDiscard()` inside this loop before calling `cgroups::freezer:freeze()` ?

Yes, I wondered about this, and actually I initially wanted to go for this approach to never interrupt of the middle of an iteration, however I chose the approach in this MR because:
- it's simpler
- I don't feel too comfortable changing this code as it's already the result of several workarounds, so as you said earlier it's not clear that any change won't break some subtle corner-case
- but more fundamentally, the reason is that if the freezing fails, then I think pretty much all bets are off. For example in case of timeout upon freeze, I'm not convinced that the call to `TaskKiller::kill` is safe because by definition some tasks are not frozen hence the kill won't be atomic and it might be possible that because of PID reuse we end up killing the wrong process. Another thing is that if freezing failed, I am not sure how reliable the thawing or if the kernel might just discard it which will also leave the cgroup in a bad state.

However I'm open to trying it if you prefer the non-interruption approach.

It's so sad that there is no simple and reliable way to do this. AFAICT systemd doesn't use the freezer cgroup because of its various issues - it actually suffers from PID reuse race conditions, looks like they are considering using freezer in cgroupsv2: https://github.com/systemd/systemd/issues/13101

And sadly, while PID namespace mentions in the various comments sounds promising, the problem is that it could definitely break some tasks so it's not like we can just switch to using it.
 
> Btw, how are you testing the code handling freeze/kill failures?
> In my experience, the simplest way to reliably create a process in an uninterruptable sleep (D state) is to create a file on some FUSE-backed FS (say, sshfs), mmap that file, stop the program backing the FS with SIGSTOP, and repeatedly read/write the mmap-ed bytes until the process eventually hangs. That's rather complicated, takes 30 seconds of read/write on my kernel and is not really usable in tests... I'm wondering if there is a simpler option.

It depends.
If the goal is really to create a process in uninterruptible sleep, then yes I'm not aware of any way much simpler - typically I'd use a write to an hard NFS mount, and use e.g. iptables to drop the packets and cause the process to hang.

However I don't believe it's quite necessary to test the freezer code.
For this change, I just re-ran some cgroups related tests in a loop because it was enough to trigger the issue.

For the previous changes I've made to the freezer code - some workarounds for kernel bugs - I'd generally just use `strace --inject` to either introduce faults or delays, e.g.:
```
cf@thinkpad:~$ strace -ttT --inject=read:delay_enter=1s /bin/cat /etc/fstab >/dev/null 
[...]
20:34:21.932568 openat(AT_FDCWD, "/etc/fstab", O_RDONLY) = 3 <0.000041>
[...]
20:34:21.932936 read(3, "# /etc/fstab: static file system"..., 131072) = 788 <1.000289>
20:34:22.933402 write(1, "# /etc/fstab: static file system"..., 788) = 788 <0.000037>
20:34:22.933580 read(3, "", 131072)     = 0 <1.000115>
20:34:23.933857 munmap(0x7ff03b46b000, 139264) = 0 <0.000074>
[...]
```

This can be used to e.g. cause timeouts while reading/writing to `freezer.state` file, which is enough to exercise some interesting code paths.
And I quite like it because it's very generic and can be used for many problems.

For something purely freezer-specific, another way I can think of to trigger freezer errors, and which can actually probably be done as part of an automated test, is to use nested cgroups:
```
root@thinkpad:~# # create parent cgroup
root@thinkpad:~# mkdir /sys/fs/cgroup/freezer/parent
root@thinkpad:~# # create child cgroup
root@thinkpad:~# mkdir /sys/fs/cgroup/freezer/parent/child
root@thinkpad:~# # freeze parent - since freezing is recursive, the child will be frozen
root@thinkpad:~# echo FROZEN > /sys/fs/cgroup/freezer/parent/freezer.state
root@thinkpad:~# cat /sys/fs/cgroup/freezer/parent/freezer.state 
FROZEN
root@thinkpad:~# cat /sys/fs/cgroup/freezer/parent/child/freezer.state 
FROZEN
root@thinkpad:~# # attempt to thaw the child
root@thinkpad:~# echo THAWED > /sys/fs/cgroup/freezer/parent/child/freezer.state 
root@thinkpad:~# # not working since parent is still frozen
root@thinkpad:~# cat /sys/fs/cgroup/freezer/parent/child/freezer.state 
FROZEN
root@thinkpad:~# cat /sys/fs/cgroup/freezer/parent/freezer.state 
FROZEN
root@thinkpad:~# # thaw parent, and child thaws as well
root@thinkpad:~# echo THAWED > /sys/fs/cgroup/freezer/parent/freezer.state 
root@thinkpad:~# cat /sys/fs/cgroup/freezer/parent/freezer.state 
THAWED
root@thinkpad:~# cat /sys/fs/cgroup/freezer/parent/child/freezer.state 
THAWED
root@thinkpad:~# 
```

abudnik:
IIRC, Mesos leaves a task in the KILLING state if `cgroup::destroy()` fails or hangs. The agent can't force the kernel to deallocate resources in the cgroup, so it leaves the task in that state to prevent resource leaks. Otherwise, the agent would eventually allocate all resources of the host on a buggy kernel.

I remember a few jira tickets describing similar issues. Maybe it would make more sense to add some links as there have been investigations some time ago.

qianzhangxa:
@cf-natali Can you please let me know your kernel version and which test has the most possibility to fail due to this issue? I'd like to try to reproduce it in my machine.

cf-natali:
Running the following usually reproduces the problem after a couple iterations:

```
root@thinkpad:/home/cf/src/mesos/build# ./bin/mesos-tests.sh --gtest_filter="SlaveRecoveryTest/0.ExecutorDanglingLatestSymlink" --gtest_repeat=-1 --gtest_break_on_failure                                         
```

```
root@thinkpad:~# uname -a
Linux thinkpad 5.10.0-7-amd64 #1 SMP Debian 5.10.40-1 (2021-05-28) x86_64 GNU/Linux
```

I'll try to update the MR with the new approach soon.

cf-natali:
@qianzhangxa Did you manage to reproduce it?

The updated MR only handles the discard at the beginning of a freeze/kill/thaw cycle, therefore ensuring we don't leave a cgroup frozen.
Obviously it means that discard will be ignored while the kill/thaw is in progress, but I think it's what we want since interrupting them would be quite bad, and in practice kill is non-blocking and thaw shouldn't block forever.

src/linux/cgroups.cpp
```diff
@@ -1425,16 +1424,15 @@ class TasksKiller : public Process<TasksKiller>
...
-    // pending signals to be delivered. See MESOS-1689 for details.
-    // This is a short term hack until we have PID namespace support.
-    return Future<bool>(true)
+    // Wait until the freeze is cancelled, and then attempt to kill the
+    // processes before we thaw again, due to a bug in the kernel. See
+    // MESOS-1758 for more details.  We thaw the cgroup before trying to freeze
+    // again to allow any pending signals to be delivered. See MESOS-1689 for
+    // details.  This is a short term hack until we have PID namespace support.
+    return future
+      .recover([](const Future<Nothing>&){ return Future<Nothing>(Nothing()); })
```
qianzhangxa, line 1435: I think you intention is to wait until `future` is in `DISCARDED` state before starting retry, right? But if that happens, will we just return from here without calling the 3 methods below (i.e. no retry)?

qianzhangxa:
@cf-natali I did not upgrade the kernel of my machine yet (5.10.0 seems a bit too new for me). Thanks for updating the PR, I left a comment.

src/linux/cgroups.cpp
cf-natali: Hm, no?
We won't return, we carry on with the 3 methods below. I confirmed it by looking at logs.
The `[](const Future<Nothing>&){ return Future<Nothing>(Nothing()); }` is basically just a no-op in order to be able to call `recover`.
Is there a simpler way?

qianzhangxa:
> Which means that the task killer can be interrupted after it's frozen the cgroup, but before killing the tasks and thawing.

@cf-natali I am a bit confused about this, if the cgroup is frozen, then I think we should see a log message like `Successfully froze cgroup ...`, right? But in the logs you pasted (see below), I do not see such message, it just tried to thaw the cgroup even the cgroup was not frozen. I am not sure why it can happen.

```
I0516 23:58:42.638830 466897 linux_launcher.cpp:606] Destroying cgroup '/sys/fs/cgroup/freezer/mesos_test_98d2814b-03ad-4c14-a3f9-832ab688e386/3139257f-4e7a-4828-9af4-75c867846558'
I0516 23:58:42.639083 466898 composing.cpp:343] Finished recovering all containerizers
I0516 23:58:42.639549 466898 cgroups.cpp:2934] Freezing cgroup /sys/fs/cgroup/freezer/mesos_test_98d2814b-03ad-4c14-a3f9-832ab688e386/3139257f-4e7a-4828-9af4-75c867846558
I0516 23:58:42.714794 466899 cgroups.cpp:2952] Thawing cgroup /sys/fs/cgroup/freezer/mesos_test_98d2814b-03ad-4c14-a3f9-832ab688e386/3139257f-4e7a-4828-9af4-75c867846558
```

cf-natali:
> > Which means that the task killer can be interrupted after it's frozen the cgroup, but before killing the tasks and thawing.
> 
> @cf-natali I am a bit confused about this, if the cgroup is frozen, then I think we should see a log message like `Successfully froze cgroup ...`, right? But in the logs you pasted (see below), I do not see such message, it just tried to thaw the cgroup even the cgroup was not frozen. I am not sure why it can happen.
> 
> ```
> I0516 23:58:42.638830 466897 linux_launcher.cpp:606] Destroying cgroup '/sys/fs/cgroup/freezer/mesos_test_98d2814b-03ad-4c14-a3f9-832ab688e386/3139257f-4e7a-4828-9af4-75c867846558'
> I0516 23:58:42.639083 466898 composing.cpp:343] Finished recovering all containerizers
> I0516 23:58:42.639549 466898 cgroups.cpp:2934] Freezing cgroup /sys/fs/cgroup/freezer/mesos_test_98d2814b-03ad-4c14-a3f9-832ab688e386/3139257f-4e7a-4828-9af4-75c867846558
> I0516 23:58:42.714794 466899 cgroups.cpp:2952] Thawing cgroup /sys/fs/cgroup/freezer/mesos_test_98d2814b-03ad-4c14-a3f9-832ab688e386/3139257f-4e7a-4828-9af4-75c867846558
> ```

It happens in case of freeze timeout: https://github.com/apache/mesos/blob/master/src/linux/cgroups.cpp#L1423

The freeze is discarded, then it calls `kill` - which doesn't log anything - then it calls `thaw`.

qianzhangxa:
> It happens in case of freeze timeout: https://github.com/apache/mesos/blob/master/src/linux/cgroups.cpp#L1423
> The freeze is discarded, then it calls kill - which doesn't log anything - then it calls thaw.

I see the timeout for freeze is 10 seconds (`FREEZE_RETRY_INTERVAL`), so if that occurs, there should be ~10s interval between the log messages `Freezing cgroup ...` and `Thawing cgroup ...`, but that's not the case in the log messages you posted:
```
I0516 23:58:42.638830 466897 linux_launcher.cpp:606] Destroying cgroup '/sys/fs/cgroup/freezer/mesos_test_98d2814b-03ad-4c14-a3f9-832ab688e386/3139257f-4e7a-4828-9af4-75c867846558'
I0516 23:58:42.639083 466898 composing.cpp:343] Finished recovering all containerizers
I0516 23:58:42.639549 466898 cgroups.cpp:2934] Freezing cgroup /sys/fs/cgroup/freezer/mesos_test_98d2814b-03ad-4c14-a3f9-832ab688e386/3139257f-4e7a-4828-9af4-75c867846558
I0516 23:58:42.714794 466899 cgroups.cpp:2952] Thawing cgroup /sys/fs/cgroup/freezer/mesos_test_98d2814b-03ad-4c14-a3f9-832ab688e386/3139257f-4e7a-4828-9af4-75c867846558
```

qianzhangxa:
I do not have any comments for the code changes in this PR, @asekretenko Do you have further comments?

cf-natali:
> > It happens in case of freeze timeout: https://github.com/apache/mesos/blob/master/src/linux/cgroups.cpp#L1423
> > The freeze is discarded, then it calls kill - which doesn't log anything - then it calls thaw.
> 
> I see the timeout for freeze is 10 seconds (`FREEZE_RETRY_INTERVAL`), so if that occurs, there should be ~10s interval between the log messages `Freezing cgroup ...` and `Thawing cgroup ...`, but that's not the case in the log messages you posted:
> 
> ```
> I0516 23:58:42.638830 466897 linux_launcher.cpp:606] Destroying cgroup '/sys/fs/cgroup/freezer/mesos_test_98d2814b-03ad-4c14-a3f9-832ab688e386/3139257f-4e7a-4828-9af4-75c867846558'
> I0516 23:58:42.639083 466898 composing.cpp:343] Finished recovering all containerizers
> I0516 23:58:42.639549 466898 cgroups.cpp:2934] Freezing cgroup /sys/fs/cgroup/freezer/mesos_test_98d2814b-03ad-4c14-a3f9-832ab688e386/3139257f-4e7a-4828-9af4-75c867846558
> I0516 23:58:42.714794 466899 cgroups.cpp:2952] Thawing cgroup /sys/fs/cgroup/freezer/mesos_test_98d2814b-03ad-4c14-a3f9-832ab688e386/3139257f-4e7a-4828-9af4-75c867846558
> ```

That's because the test uses `Clock::pause`, `Clock::advance` and `Clock::settle` to advance the time, hence the freeze timeout occurs really quickly.
That's probably why this test is the most effective one at reproducing the issue :).

src/linux/cgroups.cpp
qianzhangxa: > We won't return, we carry on with the 3 methods below.

I am not sure about it. For example, if [these code](https://github.com/apache/mesos/blob/1.11.0/3rdparty/libprocess/src/http.cpp#L1883:L1891) get called, will the `then` below it still be called?

cf-natali: Yes, I would assume so.

cf-natali: See this example:
```
cf@thinkpad:~/src/test/mesos$ cat test_process.cpp 
#include <process/process.hpp>

using namespace process;

int
main(int argc, char* argv[])
{
  auto promise = Promise<Nothing>();
  auto future = promise.future();

  future
    .recover([](const Future<Nothing>& future) -> Future<Nothing> {
      std::cout << "recover" << std::endl;
      return Nothing();
    })
    .then([](const Future<Nothing>& future) {
      std::cout << "then" << std::endl;
      return Nothing();
    });

  promise.discard();
  future.await();
}
cf@thinkpad:~/src/test/mesos$ /tmp/test_process 
recover
then
cf@thinkpad:~/src/test/mesos$ 
```

cf-natali:
Hey,

Anything preventing to merge this?

qianzhangxa:
Sorry for the late response, I have merged it, thanks for your contributions @cf-natali!

PR #391, opened 2021-06-01
Fixed parsing of `perf` output on some locales.

cf-natali:
If the locale is such that `LC_NUMERIC` uses the comma ',' as decimal
separator, parsing won't work - because of unexpected number of fields
and floating points format - so make sure it's set to `C`.

Example:
```
[ RUN      ]
CgroupsAnyHierarchyWithPerfEventTest.ROOT_CGROUPS_PERF_PerfTest
../../src/tests/containerizer/cgroups_tests.cpp:1024: Failure
(statistics).failure(): Failed to parse perf sample: Failed to parse
perf sample line
'6376827291,,cycles,mesos_test,2011741096,100,00,3,GHz': Unexpected
number of fields (9)
[  FAILED  ]
CgroupsAnyHierarchyWithPerfEventTest.ROOT_CGROUPS_PERF_PerfTest (2157
ms)
```

Standalone reproducer, using '/' as separator for readability:
```
root@thinkpad:~# LC_NUMERIC=fr_FR.UTF-8 perf stat --field-separator "/"
-- true
0,31/msec/task-clock/306721/100,00/0/CPUs utilized
0//context-switches/306721/100,00/0/K/sec
0//cpu-migrations/306721/100,00/0/K/sec
44//page-faults/306721/100,00/0/M/sec
788234//cycles/311478/100,00/2/GHz
538077//instructions/311478/100,00/0/insn per cycle
106749//branches/311478/100,00/348/M/sec
4556//branch-misses/311478/100,00/4/of all branches
```

### DIFF
--- a/src/linux/perf.cpp
+++ b/src/linux/perf.cpp
@@ -125,6 +125,12 @@ protected:
 private:
   void execute()
   {
+    // If the locale is such that `LC_NUMERIC` uses the comma ',' as decimal
+    // separator, parsing won't work - because of unexpected number of fields
+    // and floating points format - so make sure it's set to `C`.
+    std::map<string, string> env = os::environment();
+    env["LC_ALL"] = "C";
+
     // NOTE: The supervisor childhook places perf in its own process group
     // and will kill the perf process when the parent dies.
     Try<Subprocess> _perf = subprocess(
@@ -134,7 +140,7 @@ private:
         Subprocess::PIPE(),
         Subprocess::PIPE(),
         nullptr,
-        None(),
+        env,
         None(),
         {},
         {Subprocess::ChildHook::SUPERVISOR()});


cf-natali:
@asekretenko @qianzhangxa 

It's really a corner case but it's causing one of the tests to fail on my box, and I'm trying to get them all to pass...

src/linux/perf.cpp
```diff
@@ -125,6 +125,12 @@ class Perf : public Process<Perf>
 private:
   void execute()
   {
+    // If the locale is such that `LC_NUMERIC` uses the comma ',' as decimal
+    // separator, parsing won't work - because of unexpected number of fields
+    // and floating points format - so make sure it's set to `C`.
+    std::map<string, string> env = os::environment();
+    env["LC_ALL"] = "C";
```
qianzhangxa, line 132: So here we set `LC_ALL` to `C` to make sure the output of `perf stat` always has `.` as the decimal
separator (e.g., `100.00` instead of `100,00`) regardless user's locale, right?

qianzhangxa:
> root@thinkpad:~# LC_NUMERIC=fr_FR.UTF-8 perf stat --field-separator "/"
> -- true
> 0,31/msec/task-clock/306721/100,00/0/CPUs utilized
> 0//context-switches/306721/100,00/0/K/sec
> 0//cpu-migrations/306721/100,00/0/K/sec
> 44//page-faults/306721/100,00/0/M/sec
> 788234//cycles/311478/100,00/2/GHz
> 538077//instructions/311478/100,00/0/insn per cycle
> 106749//branches/311478/100,00/348/M/sec
> 4556//branch-misses/311478/100,00/4/of all branches

When I run `LC_NUMERIC=fr_FR.UTF-8 perf stat --field-separator "/" -- true` in my machine, I see something like:
`0//cpu-migrations/779898/100.00/0.000/K/sec` which seems different from yours, it seems still using `.` as decimal separator (i.e. `100.00` and `0.000`).

src/linux/perf.cpp
cf-natali: Yep.

cf-natali:
> When I run `LC_NUMERIC=fr_FR.UTF-8 perf stat --field-separator "/" -- true` in my machine, I see something like:
> `0//cpu-migrations/779898/100.00/0.000/K/sec` which is different from yours, it seems still using `.` as decimal separator (i.e. `100.00` and `0.000`).

Might be because `LC_ALL` is set on your machine - try running:
```
# LC_ALL=fr_FR.UTF-8 perf stat --field-separator "/" -- true
```

instead?

qianzhangxa:
> It's also possible you don't have the `fr_FR.UTF-8` locale generated, in which case it wouldn't work.

Yeah, it's because that I did not have the `fr_FR.UTF-8` locale generated, now it works as expected. Thank you!

BTW, I see in the output of your test, the number of tokens is 7 (e.g. `788234//cycles/311478/100,00/2/GHz`) which should fail in [this code](https://github.com/apache/mesos/blob/1.11.0/src/linux/perf.cpp#L364:L400) since there is no case 7, right?

cf-natali:
> BTW, I see in the output of your test, the number of tokens is 7 (e.g. 788234//cycles/311478/100,00/2/GHz) which should fail in this code since there is no case 7, right?

Yes this example wouldn't work anyway but that's because it doesn't specify the cgroup, whereas the actual code does (https://github.com/apache/mesos/blob/50b729904e4b4ed7964cd5f2ac8177bca5a5a72c/src/linux/perf.cpp#L288):
```
root@thinkpad:~# perf stat --field-separator "/" --all-cpus -e cycles -- true
1987580//cycles/3437534/100.00//
```

```
root@thinkpad:~# mkdir /sys/fs/cgroup/perf_event/test
root@thinkpad:~# perf stat --field-separator "/" --all-cpus -e cycles --cgroup test -- true
<not counted>//cycles/test/0/100.00//
root@thinkpad:~# 
```

qianzhangxa:
> Yes this example wouldn't work anyway but that's because it doesn't specify the cgroup, whereas the actual code does (

Aha, I see `perf::sample` will be a no-op if no cgroup is specified:
```
Future<hashmap<string, mesos::PerfStatistics>> sample(
    const set<string>& events,
    const set<string>& cgroups,
    const Duration& duration)
{
  // Is this a no-op?
  if (cgroups.empty()) {
    return hashmap<string, mesos::PerfStatistics>();
  }

...
```

qianzhangxa (approved):

PR #393, opened 2021-06-06
Backported a boost-mpl commit to ignore GCC's -Wparentheses.

cf-natali:
In order to allow us to build with `-Werror`.
Since boost is quite a large dependency and the change is so small, it
felt much simpler and less risky than updating boost just to silence
this warning.

See https://github.com/boostorg/mpl/pull/34 for the upstream fix.

This together with https://github.com/apache/mesos/pull/392 allows to build with `-Werror` using this somewhat recent gcc:
```
gcc (Debian 10.2.1-6) 10.2.1 20210110
```

@asekretenko @qianzhangxa

### DIFF
--- a/3rdparty/boost-1.65.0.patch
+++ b/3rdparty/boost-1.65.0.patch
@@ -16,3 +16,49 @@ index c0557de71..c06eaea45 100644
 -#     pragma message("Unknown compiler version - please run the configure tests and report the results")
 -#  endif
 -#endif
+From f48fd09d021db9a28bd7b8452c175897e1af4485 Mon Sep 17 00:00:00 2001
+From: Romain Geissler <romain.geissler@amadeus.com>
+Date: Thu, 1 Mar 2018 09:59:33 +0100
+Subject: [PATCH] Ignore gcc 8 warnings.
+
+---
+ boost/mpl/assert.hpp | 15 +++++++++++++--
+ 1 file changed, 13 insertions(+), 2 deletions(-)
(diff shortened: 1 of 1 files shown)


3rdparty/boost-1.65.0.patch
```diff
@@ -16,3 +16,49 @@ index c0557de71..c06eaea45 100644
...
+ 
++#if defined(BOOST_GCC) && BOOST_GCC >= 80000
++#define BOOST_MPL_IGNORE_PARENTHESES_WARNING
++#pragma GCC diagnostic push
++#pragma GCC diagnostic ignored "-Wparentheses"
++#endif
++
+ template< typename Pred >
+-failed ************ (Pred::************ 
++failed ************ (Pred::************
```
qianzhangxa, line 44: I do not quite understand this change, these two lines are identical, right? So I guess we do not need to do this change?

cf-natali: Hey @qianzhangxa ,

If you look closely the change removed a trailing space ` `.
It's a bit unrelated to the fix but this is exactly the commit which was merged upstream - https://github.com/boostorg/mpl/commit/3c1910797cf2a9e6f18a0a929bec95837accfb9e - so better to keep it as-is.

Cheers,

qianzhangxa: Yeah, you are right, sorry for my bad.

PR #394, opened 2021-06-23
[linux] <ldcache.parse> fixed to handle excess tail data in Ubuntu <ld.so.cache>.

surahman:
**_[MESOS-10244](https://issues.apache.org/jira/browse/MESOS-10224)_**

There is excess data on the tail end of the `ld.cache.so` file on `Ubuntu 21.04`. With this fix the `data` pointer check will not fail if it falls short of the end of the cache's buffer end.

### DIFF
--- a/src/linux/ldcache.cpp
+++ b/src/linux/ldcache.cpp
@@ -222,16 +222,16 @@ Try<vector<Entry>> parse(const string& path)
 
   // Adjust the pointer to add on the additional size of the strings
   // contained in the string table. At this point, 'data' should
-  // point to an address just beyond the end of the file.
+  // point to an address just inside or beyond the end of the file.
   data += headerNew->stringsLength;
-  if ((size_t)(data - buffer->data()) != buffer->size()) {
+  if ((size_t)(data - buffer->data()) > buffer->size()) {
     return Error("Invalid format");
   }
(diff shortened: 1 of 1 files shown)


src/linux/ldcache.cpp
```diff
@@ -224,7 +224,7 @@ Try<vector<Entry>> parse(const string& path)
   // contained in the string table. At this point, 'data' should
   // point to an address just beyond the end of the file.
```
cf-natali, line 225: I think the comment is stale, since `data` won't necessarily point just beyond the end of the file anymore.

Similarly, the comment below will also be stale:
https://github.com/apache/mesos/blob/15c493c4dba2bf7b5db514e1daefc9c8d4f11bf6/src/linux/ldcache.cpp#L231

```
  // Make sure the very last character in the buffer is a '\0'.
```

cf-natali:
LGTM - @qianzhangxa can you review and merge?

qianzhangxa:
> LGTM - @qianzhangxa can you review and merge?

Sure. Thanks @surahman for your contribution!

surahman:
@cf-natali and @qianzhangxa thank you, I am looking forward to making more substantial contributions going forward.

PR #398, opened 2021-06-29
MESOS-10225: Update documentation to systemd's parameter delegate

andreaspeters:
With these PR I would like to add the systemd parameter "delegate" into the documentation of the mesos-agent. 
See Issue: https://issues.apache.org/jira/browse/MESOS-10225

Therefore these PR would not remove our asf.yaml file (https://github.com/apache/mesos-site/pull/2), I also add it into the middleman build process. I know, normally it should be two PR's.

### DIFF
--- a/docs/mesos-containerizer.md
+++ b/docs/mesos-containerizer.md
@@ -66,3 +66,14 @@ Mesos supports the following built-in isolators.
 - [volume/secret](secrets.md#file-based-secrets)
 - [windows/cpu](isolators/windows.md#cpu-limits)
 - [windows/mem](isolators/windows.md#memory-limits)
+
+## Systemd Integration
+
+To prevent systemd from manipulating cgroups managed by the agent, 
+it's recommended to add 'Delegate' under 'Service' in the service 
+unit file of Mesos agent, for example:
+<pre><code>
+[Service]
+Delegate=true
+
+</code></pre>
--- a/support/packaging/common/mesos-slave.service
+++ b/support/packaging/common/mesos-slave.service
@@ -11,6 +11,7 @@ RestartSec=20
 LimitNOFILE=16384
 CPUAccounting=true
 MemoryAccounting=true
+Delegate=true
 
 [Install]
 WantedBy=multi-user.target


qianzhangxa:
@andreaspeters The changes for persisting asf.yaml configuration seems not related to MESOS-10225, right? Could you please create a dedicated PR for that?

andreaspeters:
No problem, but then it's important to merge the (new) asf PR first and then these PR therefore the jenkins websitebot will not remove the current .asf.yaml. :-) Is that ok for you?

qianzhangxa:
> No problem, but then it's important to merge the (new) asf PR first and then these PR therefore the jenkins websitebot will not remove the current .asf.yaml. :-) Is that ok for you?

Do you mean merge the new asf.yaml PR before merging this PR? I do not quite follow, I think they are totally separate PRs, right? Why is there such a dependency?

docs/configuration/agent.md
```diff
@@ -1854,5 +1854,33 @@ quotas for container sandbox directories. Valid project IDs range from
 1 to max(uint32). (default `[5000-10000]`)
 </td>
 </tr>
+</table>
+
+## Systemd Integration
```
qianzhangxa, line 1859: Hmm,  I think this file may not be the proper place to add this section since this file is all about Mesos agent flags but nothing else. Maybe we should add this section into https://github.com/apache/mesos/blob/master/docs/mesos-containerizer.md?

andreaspeters:
Yes thats what I mean. And normally u are right, it's should be two PR's. 
So, the relation between both is the website. If we change the docs, that will also change the website and jenkins will generate and publish it. The website generation is doing by middleman and if it see the current asf.yaml then he will removed it because it's not a part of his configuration.

docs/configuration/agent.md
andreaspeters: Thats a good point. Actually I'm not 100% happy with both. It have to be where it's more visible if the user have a problem and start to read the docs. I think, the user will have a look in the docs of the agents first. :man_shrugging:

qianzhangxa: I see your point. But `agent.md` is only for documenting agent flags, and `Delegate=yes` is not an agent flags.

andreaspeters:
Ok, I open a seperated PR for the asf.yaml. :-)  https://github.com/apache/mesos/pull/399

site/Gemfile.lock
```diff
@@ -161,4 +161,4 @@ DEPENDENCIES
   tzinfo-data
 
 BUNDLED WITH
-   1.17.2
+   2.2.21
```
qianzhangxa, line 164: Could you please comment why we need this change?

qianzhangxa (commented):
> Ok, I open a seperated PR for the asf.yaml. :-) #399

Thank you!

site/Gemfile.lock
andreaspeters: It happend after the build.sh. :man_shrugging: I can change to the old version. What do u think?

docs/configuration/agent.md
andreaspeters: ok, then I move it to containerizer.md. :-)

site/Gemfile.lock
cf-natali: Yeah might be better/

support/packaging/common/mesos-slave.service
```diff
@@ -11,6 +11,7 @@ RestartSec=20
 LimitNOFILE=16384
 CPUAccounting=true
 MemoryAccounting=true
+Delegate=yes
```
cf-natali, line 14: Maybe `Delegate=true` for consistency with the others?

docs/configuration/agent.md
```diff
@@ -1854,5 +1854,33 @@ quotas for container sandbox directories. Valid project IDs range from
...
+      </th>
+      <th>
+        Explanation
+      </th>
+    </tr>
+  </thead>
 
+  <tr id="systemd_delegate">
+    <td>
+      Delegate=yes
```
cf-natali, line 1875: Same thing here?

docs/configuration/agent.md
```diff
@@ -1854,5 +1854,33 @@ quotas for container sandbox directories. Valid project IDs range from
...
+    </td>
+    <td>
+      To prevent systemd from manipulating cgroups managed by the agent, it's recommended to add 'Delegate' under 'Service'.
+      Example:
+      <pre><code>
+        [Service]
+        Delegate=yes
```
cf-natali, line 1882: And here.

site/Gemfile.lock
andreaspeters: done

docs/configuration/agent.md
andreaspeters: definitely, when then everywhere :-D

andreaspeters:
Everything solved. Later, I will build the website to have a look if everything is fine and then push the changes.

docs/configuration/agent.md
qianzhangxa: Thanks, after you move it to containerizer.md, please mark this thread as resolved.

qianzhangxa:
@andreaspeters Could you please update the commit message by removing `persist asf.yaml configuration` from there?

andreaspeters:
> @andreaspeters Could you please update the commit message by removing `persist asf.yaml configuration` from there?

I will. :+1:

andreaspeters:
I was able to build the website without errors. U can merge it @qianzhangxa if u like. :-)

docs/configuration/agent.md
```diff
@@ -1854,5 +1854,5 @@ quotas for container sandbox directories. Valid project IDs range from
 1 to max(uint32). (default `[5000-10000]`)
 </td>
 </tr>
-
 </table>
+
```
qianzhangxa, line 1858: I think we should leave this file untouched, right?

docs/mesos-containerizer.md
```diff
@@ -66,3 +66,33 @@ Mesos supports the following built-in isolators.
 - [volume/secret](secrets.md#file-based-secrets)
 - [windows/cpu](isolators/windows.md#cpu-limits)
 - [windows/mem](isolators/windows.md#memory-limits)
+
+
+## Systemd Integration
+
+<table class="table table-striped">
```
qianzhangxa, line 73: I am not sure if we need a table here, maybe just putting the following statement here should be good enough?

> To prevent systemd from manipulating cgroups managed by the agent, it's recommended to add `Delegate` under `Service`, for example:
> ```
> [Service]
> Delegate=true
> ```

qianzhangxa:
> I was able to build the website without errors. U can merge it @qianzhangxa if u like. :-)

@andreaspeters Did you mean https://github.com/apache/mesos/pull/399 is ready for merge now?

docs/mesos-containerizer.md
andreaspeters: I have no special feelings for the table. :-) If you prefer without, I will remove it. Then I can also cleanup the CR in the agent.md by the way.

qianzhangxa: Yeah, let's remove the table and cleanup the CR in the agent.md : )

qianzhangxa:
> > > I was able to build the website without errors. U can merge it @qianzhangxa if u like. :-)
> > 
> > 
> > @andreaspeters Did you mean #399 is ready for merge now?
> 
> #399 is ready, yes.

Merged it, thanks!

docs/mesos-containerizer.md
andreaspeters: I'm back after holiday :-)

docs/mesos-containerizer.md
```diff
@@ -66,3 +66,14 @@ Mesos supports the following built-in isolators.
 - [volume/secret](secrets.md#file-based-secrets)
 - [windows/cpu](isolators/windows.md#cpu-limits)
 - [windows/mem](isolators/windows.md#memory-limits)
+
+
```
qianzhangxa, line 70: I think we just need one blank line rather than two.

docs/mesos-containerizer.md
```diff
@@ -66,3 +66,14 @@ Mesos supports the following built-in isolators.
...
+## Systemd Integration
+
+To prevent systemd from manipulating cgroups managed by the agent, it's recommended to add 'Delegate' under 'Service'.
+Example:
```
qianzhangxa, line 74: Better to change
> it's recommended to add 'Delegate' under 'Service'.
> Example:

to

> it's recommended to add `Delegate` under `Service` in the service unit file of Mesos agent, for example:

docs/mesos-containerizer.md
qianzhangxa: Welcome back, I just added a few minor comments :-)

docs/mesos-containerizer.md
andreaspeters: done

docs/mesos-containerizer.md
andreaspeters: also done :-)

docs/mesos-containerizer.md
andreaspeters: Thanks :-)

cf-natali:
@andreaspeters thanks for taking care of this!

PR #400, opened 2021-07-25
LinkedHashMap: fixed handling of self-assignment.

cf-natali:
Self-assigning a LinkedHashMap i.e. `map = map` would cause the map
to be cleared.
Found with clang-tidy.

@asekretenko @qianzhangxa


qianzhangxa:
Merged, thanks @cf-natali for your contribution!

PR #402, opened 2021-08-07
Fixed NestedMesosContainerizerTest hangs on errors.

cf-natali:
Those tests would use a named pipe to synchronize with the task being
started.
The problem is that if the task fails to start, reading from the pipe
would block indefinitely, making the tests just hang.
We could update the code to use a read with a timeout, however it's a
bit fiddly and it's simpler to just use the presence as a regular file
as a barrier.

See https://issues.apache.org/jira/browse/MESOS-10226 for context.
Tested by @martin-g  

@qianzhangxa

### DIFF
--- a/src/tests/containerizer/nested_mesos_containerizer_tests.cpp
+++ b/src/tests/containerizer/nested_mesos_containerizer_tests.cpp
@@ -157,6 +157,23 @@ protected:
 
     return containerConfig;
   }
+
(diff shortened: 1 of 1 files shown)


martin-g (approved):

qianzhangxa (approved):

qianzhangxa:
I have merged this PR, thanks @cf-natali for your contributions!

PR #405, opened 2021-08-16
Add framework plugin

andreaspeters:
As I mentioned, I add a framework plugin to mesos-cli and some new features for the already existing task plugin.

The features are:

- show all framework
- inspect frameworks and tasks
- disable the limitation of tasks in the task list

### DIFF
--- a/src/python/cli_new/lib/cli/http.py
+++ b/src/python/cli_new/lib/cli/http.py
@@ -34,9 +34,9 @@ def read_endpoint(addr, endpoint, config, query=None):
     """
     Read the specified endpoint and return the results.
     """
-
     try:
         addr = cli.util.sanitize_address(addr)
+
     except Exception as exception:
         raise CLIException("Unable to sanitize address '{addr}': {error}"
                            .format(addr=addr, error=str(exception)))
@@ -64,13 +64,11 @@ def read_endpoint(addr, endpoint, config, query=None):
                            .format(url=url, error=str(exception)))
 
 
-def get_json(addr, endpoint, config, condition=None, query=None):
+def get_json(addr, endpoint, config, query=None):
     """
-    Return the contents of the 'endpoint' at 'addr' as JSON data
-    subject to the condition specified in 'condition'. If we are
-    unable to read the data we throw an error.
+    Return the contents of the 'endpoint' at 'addr' as JSON data. 
+    If we are unable to read the data we throw an error.
     """
-
     data = read_endpoint(addr, endpoint, config, query)
 
     try:
@@ -78,11 +76,5 @@ def get_json(addr, endpoint, config, condition=None, query=None):
     except Exception as exception:
         raise CLIException("Could not load JSON from '{data}': {error}"
                            .format(data=data, error=str(exception)))
-
-    if not condition:
-        return data
-
-    if condition(data):
-        return data
 
     return data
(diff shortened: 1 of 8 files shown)


cf-natali:
Hey,

I'm on holiday so I'll have a look when I'm back, unless someone else beats me to it.

andreaspeters:
:joy: Take your time and enjoy your holiday.

src/python/cli_new/lib/cli/http.py
```diff
@@ -37,6 +37,7 @@ def read_endpoint(addr, endpoint, config, query=None):
 
     try:
         addr = cli.util.sanitize_address(addr)
+
```
cf-natali, line 40: Nit: why the newline?

src/python/cli_new/lib/cli/http.py
```diff
@@ -64,7 +65,7 @@ def read_endpoint(addr, endpoint, config, query=None):
                            .format(url=url, error=str(exception)))
 
 
-def get_json(addr, endpoint, config, condition=None, query=None):
+def get_json(addr, endpoint, config, query=None):
```
cf-natali, line 68: Are you sure that nothing uses `condition`?

Also, if you remove t, please remove the mention in the docstring as well:
```
subject to the condition specified in 'condition'. If we are
```

cf-natali (commented):
I think it would be better to split the changes to the task plugin into its own MR.

andreaspeters:
> I think it would be better to split the changes to the task plugin into its own MR.

Oh please don't push me to split it. :-( Thats very painfull. There will be more PR's for the mesos-cli in the future. Then I will keep an eye to separate it.

src/python/cli_new/lib/cli/http.py
andreaspeters: No reason. :man_shrugging:  will remove it.

src/python/cli_new/lib/cli/http.py
andreaspeters: Not the mesos_cli. I had a look. But of course, I will also feel better u you keep an eye on it. :-)

cf-natali:
> > I think it would be better to split the changes to the task plugin into its own MR.
> 
> Oh please don't push me to split it. :-( Thats very painfull. There will be more PR's for the mesos-cli in the future. Then I will keep an eye to separate it.

Really?
The problem is that fundamentally this MR contains two unrelated changes, so it'd be nice to separate.
Either to review, commit, and revert if necessary...

andreaspeters:
> > > I think it would be better to split the changes to the task plugin into its own MR.
> > 
> > 
> > Oh please don't push me to split it. :-( Thats very painfull. There will be more PR's for the mesos-cli in the future. Then I will keep an eye to separate it.
> 
> Really?
> The problem is that fundamentally this MR contains two unrelated changes, so it'd be nice to separate.
> Either to review, commit, and revert if necessary...

Charles u are killing me. :joy:  But I will do it.

andreaspeters:
Shell I open three PR's? Technically  there are three kind of changes.

1. Add Framework Plugins
2. Update Task Plugins
3. Update mesos_cli lib.

What do u think?

cf-natali:
Sounds good!

cf-natali:
@andreaspeters how's the refactoring going? :)

andreaspeters:
> @andreaspeters how's the refactoring going? :)

Didn't found the time the last week. Hopefully the next two weeks.

cf-natali:
@andreaspeters I'm closing this one since you've split it into separate MRs.

PR #409, opened 2021-09-21
Refactored the Python CLI code.

andreaspeters:
Like I promised in PR#405 (https://github.com/apache/mesos/pull/405), this is one of the three PR's I will open.  This PR will change:

- Remove unneeded conditions parameter to "filter" the json call. 
- Remove the limitation of "task list".
- Add global function to get the host address of a framework (for future use in plugins)
- Add global function to get all frameworks connected to mesos (for future use in plugins)

@cf-natali I will open the other two PR's when this one is reviewed. :-)

### DIFF
--- a/src/python/cli_new/lib/cli/http.py
+++ b/src/python/cli_new/lib/cli/http.py
@@ -64,13 +64,11 @@ def read_endpoint(addr, endpoint, config, query=None):
                            .format(url=url, error=str(exception)))
 
 
-def get_json(addr, endpoint, config, condition=None, query=None):
+def get_json(addr, endpoint, config, query=None):
     """
-    Return the contents of the 'endpoint' at 'addr' as JSON data
-    subject to the condition specified in 'condition'. If we are
-    unable to read the data we throw an error.
+    Return the contents of the 'endpoint' at 'addr' as JSON data.
+    If we are unable to read the data we throw an error.
     """
-
     data = read_endpoint(addr, endpoint, config, query)
 
     try:
@@ -78,11 +76,5 @@ def get_json(addr, endpoint, config, condition=None, query=None):
     except Exception as exception:
         raise CLIException("Could not load JSON from '{data}': {error}"
                            .format(data=data, error=str(exception)))
-
-    if not condition:
-        return data
-
-    if condition(data):
-        return data
 
     return data
--- a/src/python/cli_new/lib/cli/mesos.py
+++ b/src/python/cli_new/lib/cli/mesos.py
@@ -49,13 +49,8 @@ def get_agent_address(agent_id, master, config):
     Given a master and an agent id, return the agent address
     by checking the /slaves endpoint of the master.
     """
-    try:
-        agents = http.get_json(master, "slaves", config)["slaves"]
-    except Exception as exception:
-        raise CLIException("Could not open '/slaves'"
-                           " endpoint at '{addr}': {error}"
-                           .format(addr=master,
-                                   error=exception))
+    agents = get_agents(master, config)
+
     for agent in agents:
         if agent["id"] == agent_id:
             return agent["pid"].split("@")[1]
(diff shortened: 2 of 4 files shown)


src/python/cli_new/lib/cli/http.py
```diff
@@ -64,13 +64,11 @@ def read_endpoint(addr, endpoint, config, query=None):
                            .format(url=url, error=str(exception)))
 
 
-def get_json(addr, endpoint, config, condition=None, query=None):
+def get_json(addr, endpoint, config, query=None):
```
cf-natali, line 67: So interestingly the only call site which seems to pass 4 arguments  to `get_json` is this: https://github.com/apache/mesos/blob/c66aaf47768aa0e5b66ee9d830f634f4dd682c43/src/python/cli_new/lib/cli/tests/base.py#L280

However it's apparently passing a `timeout` argument, which doesn't seem to be accepted here.
Looking at `git blame`, I think you might have broken it here: https://github.com/apache/mesos/commit/c66aaf47768aa0e5b66ee9d830f634f4dd682c43#diff-1b933ba1cadd6e125668ca2646e6376b740919e25e905142108f9cc324bafb4a

As you can see you removed the `timeout` parameter from `get_json` but the test still uses it.

So I have two questions:
- have you run the tests, or do you know how to run them?
- we should probably fix the test as well

andreaspeters: Yes I run the test, I will have a look why it's working on my site. Because, u are right, it should fail at that point.

andreaspeters: ok, after cleanup, the test fails also on my site.  

But the problem is not timeout, it's the removed condition variable. I can re-implement it. But it's only used for the test. :man_shrugging: What do u think? Shell I let it inside or remove it and change the test functions?

andreaspeters: Well, I decide to change the test functions. If u don't feel fine with it, it's ok and I will re-implement the condition parameter. :-)

cf-natali: Thanks, I think that's fine!

src/python/cli_new/lib/cli/mesos.py
```diff
@@ -68,6 +63,35 @@ def get_agents(master, config):
...
+    Given a master and an framework id, return the framework address
+    by checking the /master/frameworks endpoint of the master.
+    """
+    frameworks = get_frameworks(master, config)
+
+    for framework in frameworks:
+        if framework["id"] == framework_id:
+            return framework["webui_url"]
+    # pylint: disable=line-too-long
+    raise CLIException("Unable to find framework '{id}'".format(id=framework_id))
```
cf-natali, line 79: Maybe better to indent on a new line than disable the pylint warning?

src/python/cli_new/lib/cli/tests/base.py
```diff
@@ -272,41 +274,33 @@ def launch(self, timeout=TIMEOUT):
...
 
-        try:
-            # pylint: disable=missing-docstring
-            def single_slave(data):
-                return len(data["slaves"]) == 1
-
-            http.get_json(self.flags["master"], "slaves", single_slave, timeout)
-        except Exception as exception:
+        data = http.get_json(self.flags["master"], "slaves", self.config)
+        if len(data["slaves"]) == 1:
```
cf-natali, line 278: I don't understand this change - it looks like before, the test only worked with a single slave.
Now, it looks like it will throw if there is a single slave?
```
if len(data["slaves"]) == 1:
```

andreaspeters: yes, it is the same. But since the get_json function does not have the condition, I have to change it a little.Before, the function single_slave() was a parameter (condition) for get_json(). And in get_json there was a call for condition. 
No matter, I think the whole test never really worked before. I had a look inside. The mesos-cli test even check if it's possible to start a task. But, the mesos-cli cannot start a task. :man_shrugging: So, I think the whole test should be rewritten. But thats nothing I will do at the moment.
What I'm doing is, I'm using mesos-cli by self at least some weeks long before my PR.

src/python/cli_new/lib/cli/mesos.py
andreaspeters: :joy:don't know why I even add it.

cf-natali:
Thanks, looking forward to the other MRs!

PR #410, opened 2021-10-01
Fix old sandboxes not being GC'ed issue

xiao66xiang:
Some old sandboxes were not GC'ed because Duration `d` became negative and caused `Timeout::in(d)` to overflow. Hence we got a big number for `removalTime` and those old sandboxes were never sent to remove.

Jira: https://issues.apache.org/jira/browse/MESOS-10232


bmahler:
While you're here, would you mind updating it to grab a single clock now value at the beginning? The double call makes it tricky to reason about the overflow logic holding true on the second calls as well.

bmahler:
One more thing, we have to have PRs split across changes to stout/libprocess/mesos, so you'd need one patch for the timeout header and one for the mesos test.

bmahler:
Looks good, will you be sending a separate PR for the mesos tests?

Can you update the commit title / description to reflect the timeout library issue and not the mesos related issue? We treat libprocess / stout as separate libraries so try to avoid talking about mesos in their commits.

xiao66xiang:
Yes I'll send a separate PR for the mesos tests.

mengzhugithub:
LGTM. @bmahler let's merge this?

@xiao66xiang please add the mesos test in a separate PR. thanks!

cf-natali:
@xiao66xiang Thanks!

PR #411, opened 2021-10-04
MESOS-10230:  update jquery to version 3.6.0.

andreaspeters:
Like in [MESOS-10230](https://issues.apache.org/jira/browse/MESOS-10230) mentioned, I've updated JQuery to version 3.6.0.


cf-natali:
Thanks!

I think we need to do the same thing for https://github.com/apache/mesos-site?

andreaspeters:
> Thanks!
> 
> I think we need to do the same thing for https://github.com/apache/mesos-site?

yes, I will open a PR for that after the mesos-cli. :-) The web-site will not have a impact of the security scanner from "p engels". :-)

PR #412, opened 2021-10-15
Add a gc test case with negative duration scheduled

xiao66xiang:
When a file is scheduled to be GC'ed with a negative duration, it should be GC'ed immediately.

Jira: https://issues.apache.org/jira/browse/MESOS-10232


cf-natali:
@xiao66xiang Thanks for the test as well!

PR #415, opened 2021-11-29
Explicitely fail content length fetch in case of error

Lqp1:
Fetching artifacts is done in two steps:
* use a HEAD request to get size of the remote object
* allocate needed space on disk (by eventually removing old cached data)
* fetch artifact
* validate size of fetched artifact and adapt cache behavior if there is no match (but it's outside the scope of this PR)

The issue takes place if the remote server does not answer properly to HEAD request (ie a 404 or a 500), the current code will return the size of the 404 or 500 page as if it was the size of the artifact.
There is no "real" issue now that #414 is merged, but there could still be some rare issues if the artifact is significantly larger and available cache is too small (Mesos would write more on disk than configured for its cache).

This patch will make Mesos fail sooner if HEAD request fails, but most of the time this is not an issue as the second request (GET) will fail too anyway.

### DIFF
--- a/3rdparty/stout/include/stout/net.hpp
+++ b/3rdparty/stout/include/stout/net.hpp
@@ -117,9 +117,16 @@ inline Try<Bytes> contentLength(const std::string& url)
   curl_easy_setopt(curl, CURLOPT_NOBODY, 1);
 
   CURLcode curlErrorCode = curl_easy_perform(curl);
-  if (curlErrorCode != 0) {
+  if (curlErrorCode != CURLE_OK) {
     curl_easy_cleanup(curl);
     return Error(curl_easy_strerror(curlErrorCode));
+  }
+
+  long response_code;
+  curl_easy_getinfo(curl, CURLINFO_RESPONSE_CODE, &response_code);
+  if(response_code < 200 || response_code >= 300) {
+      curl_easy_cleanup(curl);
+      return Error("Curl contentLength request returned non 2XX code");
   }
 
   double result;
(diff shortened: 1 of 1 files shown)


andreaspeters:
@cf-natali @qianzhangxa  can u help here?

andreaspeters:
@cf-natali for me it looks fine and with the response code handling more "nice" then before. :+1:

cf-natali:
Hey,

I need to think about this a bit ore, I'm a bit concerned about the potential non backward-compatibility.

andreaspeters:
May I asked where you see potential backward compatibility problems?

cf-natali:
One possibility is the server returning the correct `Content-Length`, but returning a non-`200` status.

It'd be a bit weird but who knows.

Also, as explained by @Lqp1, the following `GET` will fail anyway in case of error - and generally relying on `HEAD`/`Content-Length` is fragile since it's not atomic: the file could be modified between the `HEAD` and the `GET`.

So given the risk of breakage, and the limited impact since the downstream problem has been fixed, I'd be tempted to ignore.

andreaspeters:
Ok, I understand. More risk then we can win makes also no sense for me. But in these case, I think we should close these PR. :-)

Lqp1:
Hello :)
Thanks for your feedback.

> the file could be modified between the HEAD and the GET.

cf-natali:
> Just FYI, we are now running this patch in production, without any issue so far.

Yes - I'm sure it works as intended in your environment, the problem is guaranteeing it will won't regress anywhere :).

I'm closing, but thanks for the contribution!

PR #416, opened 2021-12-08
ADD: Task inspect function to the mesos-cli.

andreaspeters:
Add the possibility to inspect mesos tasks via mesos-cli. 

As example:

```
$ mesos task inspect test.instance-03850279-536a-11ec-847b-020000002252._app.8
{
    "id": "test.instance-03850279-536a-11ec-847b-020000002252._app.8",
    "name": "test",
    "framework_id": "6ed2b6d8-f787-4fee-988e-bbea271bab99-0000",
    "executor_id": "",
    "slave_id": "950dc31e-c9b3-4ce1-adca-e6f51ca64c50-S1",
    "state": "TASK_RUNNING",
    "resources": {
        "disk": 0.0,
        "mem": 50.0,
        "gpus": 0.0,
        "cpus": 0.1,
        "ports": "[31603-31603]"
    },
.
.
.
.
.
}
```

### DIFF
--- a/src/python/cli_new/lib/cli/plugins/task/main.py
+++ b/src/python/cli_new/lib/cli/plugins/task/main.py
@@ -18,6 +18,8 @@
 The task plugin.
 """
 
+import json
+
 from cli.exceptions import CLIException
 from cli.mesos import get_tasks
 from cli.plugins import PluginBase
@@ -65,7 +67,13 @@ class Task(PluginBase):
             },
             "short_help": "List all running tasks in a Mesos cluster",
             "long_help": "List all running tasks in a Mesos cluster"
-        }
+        },
+        "inspect": {
+            "arguments": ['<task_id>'],
+            "flags": {},
+            "short_help": "Return low-level information on the task",
+            "long_help": "Return low-level information on the task"
+        }        
     }
 
     def attach(self, argv):
@@ -75,12 +83,11 @@ class Task(PluginBase):
         """
         try:
             master = self.config.master()
-            config = self.config
         except Exception as exception:
             raise CLIException("Unable to get leading master address: {error}"
                                .format(error=exception))
 
-        task_io = TaskIO(master, config, argv["<task-id>"])
+        task_io = TaskIO(master, self.config, argv["<task-id>"])
         return task_io.attach(argv["--no-stdin"])
 
 
@@ -90,12 +97,11 @@ class Task(PluginBase):
         """
         try:
             master = self.config.master()
-            config = self.config
         except Exception as exception:
             raise CLIException("Unable to get leading master address: {error}"
                                .format(error=exception))
 
-        task_io = TaskIO(master, config, argv["<task-id>"])
+        task_io = TaskIO(master, self.config, argv["<task-id>"])
         return task_io.exec(argv["<command>"],
                             argv["<args>"],
                             argv["--interactive"],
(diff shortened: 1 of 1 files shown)


src/python/cli_new/lib/cli/plugins/task/main.py
```diff
@@ -65,7 +67,13 @@ class Task(PluginBase):
             },
             "short_help": "List all running tasks in a Mesos cluster",
             "long_help": "List all running tasks in a Mesos cluster"
-        }
+        },
+        "inspect": {
+            "arguments": ['<task_id>'],
+            "flags": {},
+            "short_help": "Return low-level information of the task",
```
cf-natali, line 74: Nit: I think it's "on the task".

src/python/cli_new/lib/cli/plugins/task/main.py
```diff
@@ -65,7 +67,13 @@ class Task(PluginBase):
...
+            "long_help": "Return low-level information of the task"
```
cf-natali, line 75: Nit: I think it's "on the task".

src/python/cli_new/lib/cli/plugins/task/main.py
```diff
@@ -143,3 +148,25 @@ def list(self, argv):
                                .format(error=exception))
 
         print(str(table))
+
+    def inspect(self, argv):
+        """
+        Show the low-level information of the task.
```
cf-natali, line 154: Same.

src/python/cli_new/lib/cli/plugins/task/main.py
```diff
@@ -143,3 +148,25 @@ def list(self, argv):
...
+
+        data = get_tasks(master, self.config)
+        try:
+            for task in data:
+                if task["id"] != argv["<task_id>"]:
+                    continue
+
+                print(json.dumps(task, indent=4))
+
+        except Exception as exception:
```
cf-natali, line 170: Can anything actually throw here?
I think it'd be fine to remove the except clause.

andreaspeters: U are right, should not throw failures. :-)

src/cli/mesos-ps
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
```
cf-natali, line 1: Why change to python3?

src/cli/mesos-cat
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
```
cf-natali, line 1: Same.

src/cli/mesos-scp
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
```
cf-natali, line 1: Same.

src/cli/mesos-tail
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
```
cf-natali, line 1: Same.

src/cli/mesos-ps
andreaspeters: Sorry, that should not be part of this commit. I will remove it.

src/cli/mesos-cat
andreaspeters: same

src/cli/mesos-scp
andreaspeters: Same

src/cli/mesos-tail
andreaspeters: Same :-)

src/cli/mesos-ps
andreaspeters: To explain, for testing purpose, I compiled Mesos for AlmaLinux under arm64. As quick fix I changed the shebang because of error messages during the rpmbuild process. It should never be part of this commit. Sorry for that. 

But maybe it would be interesting, Mesos is working quite well under AlmaLinux (x86_64 and arm64). :-)

cf-natali:
Thanks!

PR #419, opened 2022-02-16
Added Capability.Type.QUOTA_V2 to v1 operator API.

cf-natali:
Closes #MESOS-10235.


qianzhangxa:
LGTM, thanks @cf-natali!

PR #420, opened 2022-03-25
ADD: unofficial repository.

andreaspeters:
@qianzhangxa Like we talked, here my PR to add a unofficial community repository to provide mesos binaries.

### DIFF
--- a/site/source/downloads.html.erb
+++ b/site/source/downloads.html.erb
@@ -36,18 +36,30 @@ breadcrumb: Downloads
       </ul>
       </p>
 
-      <p>Once you have downloaded Mesos, follow the
+      <p>
+        Once you have downloaded Mesos, follow the
         <a href="/getting-started/">getting started instructions</a>,
         and take a look at the rest of the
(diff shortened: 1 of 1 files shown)


site/source/downloads.html.erb
```diff
@@ -36,18 +36,28 @@ breadcrumb: Downloads
...
         <br />
         This repository is also mirrored on
         <a href="https://github.com/apache/mesos">GitHub</a>.
       </p>
+
+      <h4>Getting community driven Mesos binaries</h4>
+      <p>
+        <i>The following repositories are unofficial and not supportet by Apache.</i>
+        <ul>
+          <li>Mesos binaries for CentOS (x86), AlmaLinux (x86, aarch64), Ubuntu (x86): <a href="https://rpm.aventer.biz">https://rpm.aventer.biz</a></li>
```
qianzhangxa, line 59: I think it is `x86_64` rather than `x86`, right?

And I'd suggest to add a newline after `<li>`.

qianzhangxa:
Thanks @andreaspeters!

PR #421, opened 2022-03-25
CHANGE: remove bintray from docs.

andreaspeters:
@qianzhangxa Like we talked, here my PR to remove bintray from the docs.

### DIFF
--- a/docs/binary-packages.md
+++ b/docs/binary-packages.md
@@ -5,30 +5,11 @@ layout: documentation
 
 # Binary Packages
 
-## Downloading the Mesos RPM
+## Downloading the Mesos binaries
 
-Download and install the latest stable RPM binary from the [Bintray Repository](https://bintray.com/apache/mesos/):
+Download and install [the latest stable Mesos binaries](https://mesos.apache.org/downloads/).
 
-    $ cat > /tmp/bintray-mesos-el.repo <<EOF
-    #bintray-mesos-el - packages by mesos from Bintray
-    [bintray-mesos-el]
-    name=bintray-mesos-el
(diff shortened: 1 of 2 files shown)


docs/binary-packages.md
```diff
@@ -5,28 +5,9 @@ layout: documentation
...
-
-    $ sudo mv /tmp/bintray-mesos-el.repo /etc/yum.repos.d/bintray-mesos-el.repo
-
-    $ sudo yum update
-
-    $ sudo yum install mesos
-
-The above instructions show how to install the latest version of Mesos for RHEL 7.
-Substitute `baseurl` the with the appropriate URL for your operating system.
+Download and install the latest stable binary: [Downloads](https://mesos.apache.org/downloads/)
```
qianzhangxa, line 10: Can we change this to the below?

> Download and install [the latest stable Mesos binaries](https://mesos.apache.org/downloads/).

docs/binary-packages.md
```diff
@@ -5,28 +5,9 @@ layout: documentation
...
 
 ## Start Mesos Master and Agent.
```
qianzhangxa, line 12: Not yours, but I think we do no need the trailing period.

andreaspeters: Sorry, I don't understand. :sweat_smile:  Do you mean we don't need "Start Mesos Master...."?

qianzhangxa: I mean it should be `## Start Mesos Master and Agent` rather than `## Start Mesos Master and Agent.`.

andreaspeters: Ah ok I got it. :-) Thanks.

qianzhangxa:
@andreaspeters Thanks for your contribution!

PR #423, opened 2022-03-28
Cleanup: git >= 2.32 does not support gitignore as symlink.

andreaspeters:
Nothing important, only a cleanup.

Git version 2.32.0 does not support gitignore as symlink ([https://github.com/git/git/blob/master/Documentation/RelNotes/2.32.0.txt](https://github.com/git/git/blob/master/Documentation/RelNotes/2.32.0.txt)). I updated setup-dev.sh to copy gitignore into the mesos root path. I also add some more filenames into gitignore. All of them was dynamically created by bootstrap.sh, setup-dev.sh and/or autoconf/make.

### DIFF
--- a/support/setup-dev.sh
+++ b/support/setup-dev.sh
@@ -27,7 +27,8 @@ __EOF__
 fi
 
 if test ! -e .gitignore; then
-  ln -s support/gitignore .gitignore
+  # Git version 2.32.0 does not support gitignore as symlink (https://github.com/git/git/blob/master/Documentation/RelNotes/2.32.0.txt)
+  cp support/gitignore .gitignore
 fi
 
 if test ! -e .reviewboardrc; then


support/setup-dev.sh
```diff
@@ -27,7 +27,7 @@ __EOF__
 fi
 
 if test ! -e .gitignore; then
```
cf-natali, line 29: Maybe add a comment explaining why a copy is necessary, pointing to https://github.com/git/git/blob/master/Documentation/RelNotes/2.32.0.txt ?

support/gitignore
```diff
@@ -39,3 +39,27 @@ CPPLINT.cfg
...
+aclocal.m4
+ar-lib
+autom4te.cache/
+build/
+centos7/
+compile
+config.guess
+config.sub
+configure
+*~
```
cf-natali, line 60: This `*~` will catch many files, are there many such files which can't be listed individually?

andreaspeters: I changed my mind, I think that does not have to be there. With that I want catch tmp files from my editor. But maybe that's to much. :sweat_smile:

support/setup-dev.sh
andreaspeters: :+1:

support/packaging/centos/build-docker-centos.sh
```diff
@@ -24,7 +24,7 @@ DOCKER_CONTEXT_DIR="${SOURCE_DIR}/centos${CENTOS_DISTRO}/rpmbuild/RPMS/x86_64"
 
 cat <<EOF > "${DOCKER_CONTEXT_DIR}/Dockerfile"
 FROM centos:${CENTOS_DISTRO}
-ADD mesos-?.?.?-*.rpm /
+ADD mesos-*.rpm /
```
cf-natali, line 27: This change seems unrelated?

support/packaging/centos/mesos.spec
```diff
@@ -1,9 +1,6 @@
-%global MESOS_VERSION
-%global MESOS_RELEASE
-
 Name:          mesos
-Version:       %{MESOS_VERSION}
-Release:       %{MESOS_RELEASE}%{?dist}
+Version:       1.11.1
```
cf-natali, line 2: This file too?

andreaspeters: Oh  no... Good that u look at it. :-)  This and the upper one does not have to be there. I will cleanup. Sorry! :-(

support/gitignore
```diff
@@ -39,3 +39,26 @@ CPPLINT.cfg
 .deps/
 .libs/
 *-stamp
+
+.clang-format
```
cf-natali, line 43: Several of those files already appear above, e.g. `clang-format`, `Makefile.in`, etc - could you make sure there are no duplicates?

Also, it would be nice to add them under the relevant sections.

cf-natali:
Hey, I suggest you only keep the `gitignore` copy in this MR, and create a new MR adding entries to `gitignore`.

cf-natali:
No worries, hope you had a nice break!

PR #431, opened 2022-04-30
Remove analytics and embeds from website

davelester:
Removes Google Analytics and remotely sourced files from the Mesos project website.


cf-natali:
Hi Dave,

I don't know anything about web development, but this seems to make sense to me.
Just one thing - I assume you tested it using the instructions available here: https://github.com/apache/mesos/blob/master/site/README.md ?

Also, I'm wondering, what motivated you to do that?
The reason I'm asking is that I can see you committed to Mesos a long time ago, and I'm curious to know if you'd be interested in helping a bit with the website since it's definitely not my area of expertise, and it could use some love.
For example many of the dependencies used to build it are really old and would need updating, see for example this recent bot PR: https://github.com/apache/mesos/pull/425

Cheers,

davelester:
Hi there! These changes are motivated by recent work that the The ASF Data Privacy team is doing to establish ASF-wide policies in this area. As part of this work, they recently published a draft policy for ASF websites: https://privacy.apache.org/policies/website-policy.html.

I has been *years* since I've contributed to Mesos, and I helped develop the original middleman-based Mesos website. Fast-forward to today: I want to make sure we're ahead of the curve when it comes to updating the site to be compliant with policy. These patches should be a solid first-pass at meeting the standards of the policy (removal of Google Analytics, social media embeds, etc).

Regarding testing: I pursued a different way to test these changes -- by running `middleman serve` locally to generate the static site. The site base template looks good after making changes in the footer and CSS/JS continue to load properly after changing the site to load them from ASF servers.

It's been a while since anyone -- including myself -- has deployed a new version of the Mesos site so my hope is that the build process still functions properly. If not, happy to work with others to help fix it.

I also see the several out-of-date Ruby libraries that we depend on. I began a separate patch to upgrade them, but for purposes of this initial contribution I wanted to focus on modifying templates and static content.

cf-natali:
Alright, let's merge it and keep an eye on the website build CI and webpage!

Looking forward to your next contribution,

cf-natali:
The CI passed: https://ci-builds.apache.org/job/Mesos/job/Mesos-Websitebot/

And the website has been updated, in case you want to double-check.


Cheers,

PR #432, opened 2022-04-30
Fixed clang-tidy warnings due to capturing this in a deferred lambda.

cf-natali:
Use `defer(self(), lambda)` instead to avoid the risk of use-after-free.

See on the mesos-tidy CI job:
```
/tmp/SRC/src/csi/v0_volume_manager.cpp:1078:13: warning: callback
capturing this should be dispatched/deferred to a specific PID
[mesos-this-capture]
      .then([=](const Map<string, string>& secrets) {
```

Together with the recent fixes merged, this fix should allow the
mesos-tidy CI job to be green again:
https://ci-builds.apache.org/job/Mesos/job/Mesos-Tidybot/

### DIFF
--- a/src/csi/v0_volume_manager.cpp
+++ b/src/csi/v0_volume_manager.cpp
@@ -1075,7 +1075,7 @@ Future<Nothing> VolumeManagerProcess::__publishVolume(const string& volumeId)
 
   if (!volumeState.node_stage_secrets().empty()) {
     rpcResult = resolveSecrets(volumeState.node_stage_secrets())
-      .then([=](const Map<string, string>& secrets) {
+      .then(process::defer(self(), [=](const Map<string, string>& secrets) {
         NodeStageVolumeRequest request_(request);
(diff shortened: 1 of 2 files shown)


src/csi/v0_volume_manager.cpp
```diff
@@ -1075,15 +1075,15 @@ Future<Nothing> VolumeManagerProcess::__publishVolume(const string& volumeId)
 
   if (!volumeState.node_stage_secrets().empty()) {
     rpcResult = resolveSecrets(volumeState.node_stage_secrets())
-      .then([=](const Map<string, string>& secrets) {
```
qianzhangxa, line 1078: Nit: Should we use `process::defer` instead? I see `process::defer` is used widely in this file.

cf-natali: Sure  - done.

qianzhangxa (approved):

PR #444, opened 2022-07-26
UPDATE: ruby version to fix bundle incompatibility.

andreaspeters:
With these PR I would like to fix the website build error. The reason why I do not update to the newest ruby version is, that we will get a lot of "deprecated" error messages. So, for now it's working, but we still have an issue here.

https://ci-builds.apache.org/job/Mesos/job/Mesos-Websitebot/1326/consoleFull

### DIFF
--- a/support/mesos-website/Dockerfile
+++ b/support/mesos-website/Dockerfile
@@ -13,14 +13,16 @@ USER root
 RUN apt-get update && \
     apt-get install -y --no-install-recommends \
       doxygen \
-      locales \
-      ruby \
-      ruby-dev \
-      rubygems && \
+      locales && \
     apt-get clean && \
     rm -rf /var/lib/apt/lists/*
 
-RUN gem install bundler
+# Install ruby version manager to get a more updated ruby version                                
+RUN curl -sSL https://rvm.io/mpapis.asc | gpg --import - && \
+    curl -sSL https://rvm.io/pkuczynski.asc | gpg --import - && \ 
+    curl -k -sSL https://get.rvm.io | bash -s stable --ruby=2.6.6 
+
+ENV PATH=/usr/local/rvm/rubies/ruby-2.6.6/bin:$PATH    
 
 RUN locale-gen en_US.UTF-8
 ENV LANG en_US.UTF-8


support/mesos-website/Dockerfile
```diff
@@ -13,14 +13,16 @@ USER root
 RUN apt-get update && \
     apt-get install -y --no-install-recommends \
       doxygen \
-      locales \
-      ruby \
-      ruby-dev \
-      rubygems && \
```
qianzhangxa, line 19: I think with `apt-get` we can also install a specific version of Ruby, right?

andreaspeters: Right, but the ruby version in the official Ubuntu 16.04 repository are to old.

cf-natali: Yes, I think it's reasonable.

We should merge this, and separately see if we can update the docker image to a more recent Ubuntu version, e g. 20.04 or 22.04.

I started working on updating other containers a couple weeks ago, it's not too much work.

andreaspeters: > and separately see if we can update the docker image to a more recent Ubuntu version, e g. 20.04 or 22.04.
:+1: definitely. Which containers do you already have on your hands? Then I can start to help with others.

cf-natali: I started looking at https://hub.docker.com/r/mesos/mesos-tidy and https://hub.docker.com/r/mesos/mesos-build

If you could update this image used for the website build that'd be great, and if you have time maybe also updating the CentOS image https://hub.docker.com/r/mesos/mesos-centos

Thanks!

andreaspeters: :+1: What do u think about to migrate from centos to almalinux? :thinking:

qianzhangxa: If you guys already started working on updating the images, then do we still need to merge this PR? I mean once the images are updated, we would need to revert the changes of this PR?

andreaspeters: Good question! Depends the mesos-build image and it's impact. If it's done and all the build jobs are still working,  then it should be very easy to update the "website build" dockerfile. What do u think @cf-natali ? First merge this PR to fix the website?

cf-natali: I think it makes sense to merge, this way we're sure the website build is fixed.
Upgrading to a new Ubuntu release might be more work, so I'd rather do it separately, taking the time.

cf-natali:
Merged, thanks Andreas!

Looking forward to updated Docker images :).

andreaspeters:
The docker image to  build "our" website, depends on the mesos-build docker images. As I see, the website build job compile mesos to create the API docs. So, I think I have to wait until you are finish with the mesos-build docker image. :sweat_smile: So, first I will update the mesos-centos image.

andreaspeters:
And thanks for the merge. :-D But it looks like there is a other issue. But I think it have nothing to do with our change.

grafik

https://ci-builds.apache.org/blue/organizations/jenkins/Mesos%2FMesos-Websitebot/detail/Mesos-Websitebot/1337/pipeline

cf-natali:
Yep indeed, I think I know what the issue is, I'll fix.

PR #446, opened 2022-08-27
Blkio isolator: handled missing CFQ statistics.

cf-natali:
CFQ was removed from 5.x kernel, so we need to silently ignore missing
CFQ statistics.

Before:
```
[----------] 1 test from CgroupsIsolatorTest [ RUN      ]
[...]
CgroupsIsolatorTest.ROOT_CGROUPS_BlkioUsage W0824 21:58:36.604790 308667
cgroups.cpp:932] Skipping resource statistic for container
b2d67073-85fa-4a0b-84a3-790791047eac because: Failed to read from
'blkio.time': No such file or directory
../../src/tests/containerizer/cgroups_isolator_tests.cpp:2656: Failure
Value of: usage->has_blkio_statistics() Actual: false Expected: true [
[...]
FAILED  ] CgroupsIsolatorTest.ROOT_CGROUPS_BlkioUsage (1888 ms)
```

### DIFF
--- a/src/slave/containerizer/mesos/isolators/cgroups/subsystems/blkio.cpp
+++ b/src/slave/containerizer/mesos/isolators/cgroups/subsystems/blkio.cpp
@@ -27,6 +27,7 @@ extern "C" {
 #include <stout/foreach.hpp>
 #include <stout/hashmap.hpp>
 
+#include "common/kernel_version.hpp"
 #include "linux/cgroups.hpp"
 
 #include "slave/containerizer/mesos/isolators/cgroups/subsystems/blkio.hpp"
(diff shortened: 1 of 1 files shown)


src/slave/containerizer/mesos/isolators/cgroups/subsystems/blkio.cpp
```diff
@@ -107,226 +107,205 @@ Future<ResourceStatistics> BlkioSubsystemProcess::usage(
   CgroupInfo::Blkio::CFQ::Statistics totalCfqRecursive;
   CgroupInfo::Blkio::Throttling::Statistics totalThrottling;
 
-  // Get CFQ statistics.
+  // CFQ was removed from kernel 5.x, so be ready to handle missing statistics.
+
   Try<vector<cgroups::blkio::Value>> time = cfq::time(hierarchy, cgroup);
```
qianzhangxa, line 112: Is there a way to detect whether CFQ is support or not? If it is not supported, then maybe we can skip calling the following `cfq::` methods.

cf-natali: Hm, not really.
The only way would be maybe to look at all block devices `/sys/block/*/queue/scheduler` and see if any of those files contains `cfq`, but I think it's more fragile.

cf-natali: OK to merge?

qianzhangxa: If we know the specific kernel version where CFQ was removed, can we use `kernelVersion()` to detect it like [this code](https://github.com/apache/mesos/blob/96339efb53f7cdf1126ead7755d2b83b435e3263/src/slave/containerizer/mesos/isolators/linux/nnp.cpp#L38)?

qianzhangxa: @cf-natali Any comments?

cf-natali: Sorry for the delay, I've been busy lately.

It was removed from kernel 5.0: https://github.com/torvalds/linux/commit/f382fb0bcef4c37dc049e9f6963e3baf204d815c

I've updated the change to explicitly change the kernel version - gitlab renders the diff in a sub-optimal way, but the change is actually very small (adding the version check).

src/slave/containerizer/mesos/isolators/cgroups/subsystems/blkio.cpp
```diff
@@ -107,230 +108,240 @@ Future<ResourceStatistics> BlkioSubsystemProcess::usage(
   CgroupInfo::Blkio::CFQ::Statistics totalCfqRecursive;
   CgroupInfo::Blkio::Throttling::Statistics totalThrottling;
 
-  // Get CFQ statistics.
-  Try<vector<cgroups::blkio::Value>> time = cfq::time(hierarchy, cgroup);
-  if (time.isError()) {
-    return Failure(time.error());
+  // Get CFQ statistics, if available - CFQ was removed from kernel 5.0.
```
qianzhangxa, line 111: Can you please add https://github.com/torvalds/linux/commit/f382fb0bcef4c37dc049e9f6963e3baf204d815c in  this comment for more context?

src/slave/containerizer/mesos/isolators/cgroups/subsystems/blkio.cpp
qianzhangxa: Thanks for the update.

Do we need to update any tests accordingly?

cf-natali: No, the test that was breaking are now passing.

src/slave/containerizer/mesos/isolators/cgroups/subsystems/blkio.cpp
cf-natali: Done.

PR #447, opened 2022-09-19
ADD: function to call endpoints with different http methods.

andreaspeters:
With these PR I would like to add the possibility to call mesos endpoints with different methods. I will use in a further PR to add  "drain agent" call into mesos-cli.


andreaspeters:
Will close this one. It does not 100% what I need.

PR #456, opened 2023-10-03
Add libnvidia-gpucomp to the list of NVIDIA driver libraries

dadap:
Upcoming versions of the NVIDIA driver will include a new component:

https://forums.developer.nvidia.com/t/new-driver-component-libnvidia-gpucomp/267060

Update the list of NVIDIA driver libraries so that it can be included in the runtime environment along with the others.


cf-natali:
Thank you!

PR #460, opened 2024-01-16
Fixed tests in PortMappingMesosTest suite.

DevinLeamy:
ROOT_NC_ContainerToContainerTCP, ROOT_NC_ContainerToContainerUDP, ROOT_NC_HostToContainerUDP, ROOT_NC_HostToContainerTCP, and ROOT_ContainerARPExternal are missing resources in ContainerConfig, which is now required after one of the upstream changes switched the isolator from using container resources specified in ExecutorInfo to using resources in ContainerConfig.

ROOT_NC_HostToContainerUDP is expecting nc to finish successfully when using invalid port. It fails with some nc implementations, e.g. Ncat from Nmap used by RHEL/CentOS, that return an error when they see ECONNREFUSED on a datagram socket.

ROOT_ContainerARPExternal tries to arping nameservers from /etc/resolv.conf. We don't have any nameservers configured in our local subnet, so we have to disable this test. This is a Twitter specific change.

ROOT_CGROUPS_RecoverMixedContainers does unnecessary clock manipulations that cause recovered task termination before its executor has a chance to re-register. It also installs a statusUpdate mock expectations that get oversaturated as they don't take TASK_STARTING update into account.

ROOT_CGROUPS_RecoverMixedKnownAndUnKnownOrphans checks a wrong status update due to a typo.

### DIFF
--- a/include/mesos/mesos.proto
+++ b/include/mesos/mesos.proto
@@ -1819,6 +1819,12 @@ message ResourceStatistics {
   // Total memory + swap usage. This is set if swap is enabled.
   optional uint64 mem_total_memsw_bytes = 37;
 
+  // Current kernel memory allocation.
+  optional uint64 mem_kmem_usage_bytes = 52;
+
+  // Current TCP buf memory allocation.
+  optional uint64 mem_kmem_tcp_usage_bytes = 53;
+
   // Hard memory limit.
   optional uint64 mem_limit_bytes = 6;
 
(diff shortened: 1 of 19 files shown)


bmahler:
Most of this patch looks good, but there is this part should be kept as a twitter-internal patch:

```
ROOT_ContainerARPExternal tries to arping nameservers from /etc/resolv.conf. We don't have any nameservers configured in our local subnet, so we have to disable this test. This is a Twitter specific change.
```

It's just one line of the diff that disables the test via `DISABLED_`.

DevinLeamy:
Closed in favor of: https://github.com/apache/mesos/pull/470

PR #461, opened 2024-01-16
Optionally scale egress bandwidth with CPU.

DevinLeamy:
Add support to isolators/port_mapping for optionally scaling egress bandwidth with CPU and with minimum and maximum limits.


bmahler:
Let's squash the extra follow-on commits into 1, I can do it prior to merging but it's best if you get used to keeping your commits organized on your end.

PR #462, opened 2024-01-16
Expose new metrics for memory usage in the cgroup.

DevinLeamy:
The metric "mem_kmem_usage_bytes" is the total kernel memory usage by
processes in the cgroup in bytes.

The metric "mem_kmem_tcp_usage_bytes" is the total memory usage for TCP
buffers in bytes.

### DIFF
--- a/include/mesos/mesos.proto
+++ b/include/mesos/mesos.proto
@@ -1819,6 +1819,12 @@ message ResourceStatistics {
   // Total memory + swap usage. This is set if swap is enabled.
   optional uint64 mem_total_memsw_bytes = 37;
 
+  // Current kernel memory allocation.
+  optional uint64 mem_kmem_usage_bytes = 52;
+
+  // Current TCP buf memory allocation.
+  optional uint64 mem_kmem_tcp_usage_bytes = 53;
+
   // Hard memory limit.
   optional uint64 mem_limit_bytes = 6;
 
--- a/include/mesos/v1/mesos.proto
+++ b/include/mesos/v1/mesos.proto
@@ -1783,6 +1783,12 @@ message ResourceStatistics {
   // Total memory + swap usage. This is set if swap is enabled.
   optional uint64 mem_total_memsw_bytes = 37;
 
+  // Current kernel memory allocation.
+  optional uint64 mem_kmem_usage_bytes = 52;
+
+  // Current TCP buf memory allocation.
+  optional uint64 mem_kmem_tcp_usage_bytes = 53;
+
   // Hard memory limit.
   optional uint64 mem_limit_bytes = 6;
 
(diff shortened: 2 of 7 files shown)


include/mesos/mesos.proto
```diff
@@ -1819,6 +1819,12 @@ message ResourceStatistics {
   // Total memory + swap usage. This is set if swap is enabled.
   optional uint64 mem_total_memsw_bytes = 37;
 
+  // Current kernel memory allocation
```
cf-natali, line 1822: Nit: other comments end with a dot '.'.

include/mesos/v1/mesos.proto
```diff
@@ -1783,6 +1783,12 @@ message ResourceStatistics {
   // Total memory + swap usage. This is set if swap is enabled.
   optional uint64 mem_total_memsw_bytes = 37;
 
+  // Current kernel memory allocation
+  optional uint64 mem_kmem_usage_bytes = 52;
```
cf-natali, line 1787: Is there a reason the valued don't match `include/mesos/mesos.proto`?

src/slave/containerizer/mesos/isolators/cgroups/subsystems/memory.cpp
```diff
@@ -460,6 +460,26 @@ Future<ResourceStatistics> MemorySubsystemProcess::usage(
...
+    result.set_mem_kmem_usage_bytes(kmem_usage.get().bytes());
+  }
+
+  Result<Bytes> kmem_tcp_usage = cgroups::memory::kmem_tcp_usage_in_bytes(
+      hierarchy, cgroup);
+
+  if (kmem_tcp_usage.isError()) {
+    return Failure("Failed to parse 'memory.kmem.tcp.usage_in_bytes': " +
+                   kmem_tcp_usage.error());
+  } else if (kmem_usage.isSome()) {
```
cf-natali, line 479: Should be `kmem_tcp_usage` here.

include/mesos/mesos.proto
DevinLeamy: Added periods.

src/slave/containerizer/mesos/isolators/cgroups/subsystems/memory.cpp
DevinLeamy: Good catch :)

include/mesos/v1/mesos.proto
DevinLeamy: There isn't a good reason. I've updated `mesos/mesos.proto` to match `mesos/v1/mesos.proto` (i.e. use the smaller field numbers).

bmahler:
Mind squashing this locally on your end since you've got follow on commits here that should be part of the original single commit?

DevinLeamy:
> Mind squashing this locally on your end since you've got follow on commits here that should be part of the original single commit?

Squashed.

PR #464, opened 2024-01-16
Add XFS quota headroom config flag.

DevinLeamy:
Introduces flag to configure XFS quota headroom.

- `xfs_quota_headroom`: The size of hard XFS quota headroom.

### DIFF
--- a/src/slave/containerizer/mesos/isolators/xfs/disk.cpp
+++ b/src/slave/containerizer/mesos/isolators/xfs/disk.cpp
@@ -214,7 +214,8 @@ Try<Isolator*> XfsDiskIsolatorProcess::create(const Flags& flags)
           quotaPolicy,
           flags.work_dir,
           totalProjectIds.get(),
-          flags.disk_watch_interval)));
+          flags.disk_watch_interval,
+          flags.xfs_quota_headroom)));
 }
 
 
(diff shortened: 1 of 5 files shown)


bmahler:
Be sure to squash the extra commit.

DevinLeamy:
> Be sure to squash the extra commit.

Squashed.

src/slave/containerizer/mesos/isolators/xfs/disk.cpp
```diff
@@ -223,14 +224,16 @@ XfsDiskIsolatorProcess::XfsDiskIsolatorProcess(
     xfs::QuotaPolicy _quotaPolicy,
     const std::string& _workDir,
     const IntervalSet<prid_t>& projectIds,
-    Duration _projectWatchInterval)
+    Duration _projectWatchInterval,
+    const Bytes& _quotaHeadroom)
```
cf-natali, line 228: Any reason to pass it by const reference instead of value?
I'd assume `Bytes`must be at most a word or two, and seems to be passed by value elsewhere.

src/tests/containerizer/xfs_quota_tests.cpp
```diff
@@ -892,6 +892,94 @@ TEST_F(ROOT_XFS_QuotaTest, VolumeUsageExceedsQuotaWithKill)
 }
 
 
+// Verify that a task that tries to consme more disk space than it has requested
```
cf-natali, line 895: Typo: `consme`.

src/slave/flags.hpp
```diff
@@ -203,6 +203,8 @@ class Flags : public virtual logging::Flags
 #if ENABLE_XFS_DISK_ISOLATOR
   std::string xfs_project_range;
   bool xfs_kill_containers;
+  bool xfs_enable_quotas;
```
cf-natali, line 206: I'm probably missing something but I can't see anything using `xfs_enable_quotas`?

DevinLeamy: There isn't any usage, good catch. The port has been removed and a new patch (https://github.com/apache/mesos/pull/475) will serve to replace the intention behind the flag (i.e. enabling the enforcement of XFS quotas).

PR #467, opened 2024-01-16
Stopped checking if destroyed cgroup is empty.

DevinLeamy:
Linux patch 9c974c77246460fa6a92c18554c3311c8c83c160 makes exiting processes that are still attached to a cgroup visible in cgroup.procs. It is a fix/workaround for the issue where the cgroup cannot be removed even though it appears empty.

We already have a workaround for the removal problem in place: cgroup destroyer retries rmdir until it is successful.

With this change we add a workaround for the new cgroup.procs behavior. We can no longer expect a cgroup to look empty after TasksKiller reaps all processes that used to run inside it. So, instead, we rely on the original workaround to ensure that the cgroup is successfuly removed and keep process listing only for debugging purposes.

Alternatively, we could await for the cgroup to become empty before successfuly returning from TasksKiller. But it appears unnecessary with the remove workaround in place.

### DIFF
--- a/src/linux/cgroups.cpp
+++ b/src/linux/cgroups.cpp
@@ -1515,17 +1515,23 @@ private:
       return;
     }
 
-    // Verify the cgroup is now empty.
+    // With the Linux patch 9c974c77246460fa6a92c18554c3311c8c83c160 it is
+    // possible that the cgroup might not be empty even though processes inside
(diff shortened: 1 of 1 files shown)


DevinLeamy:
Fixed commit message. @bmahler

PR #468, opened 2024-01-16
Fixed sorting in MountInfoTable::read.

DevinLeamy:
Current implementation of MountInfoTable sorting crashes when some parts of the mount tree are not visible in /proc/{pid}/mountinfo, which can happen, for example, in a chrooted environment. The sorter assumes that the / mount is always present and uses it as the starting node for the DFS traversal.

Additionally, this sorting algorithm loses mounts that are parents of themselves from the sorted MountInfoTable. The bug is not caught by FsTest.MountInfoTableReadSortedParentOfSelf test, so it needs to be updated as well.

The new sorting implementation supports multiple starting nodes and works in situations where the visible mount tree is disconnected. It also changes the resulting order to BFS, but that shouldn't matter because the invariant is that all parent entries appear before their child entries.

### DIFF
--- a/src/linux/fs.cpp
+++ b/src/linux/fs.cpp
@@ -193,54 +193,44 @@ Try<MountInfoTable> MountInfoTable::read(
   // them according to the invariant that all parent entries
   // appear before their child entries.
   if (hierarchicalSort) {
-    Option<int> rootParentId = None();
+    // Build an ID to entry and a parent to children IDs map for quick
+    // lookup.
+    hashmap<int, MountInfoTable::Entry> entries;
+    hashmap<int, vector<int>> children;
+    for (const MountInfoTable::Entry& ent : table.entries) {
+      entries.emplace(ent.id, ent);
 
-    // Construct a representation of the mount hierarchy using a hashmap.
-    hashmap<int, vector<MountInfoTable::Entry>> parentToChildren;
-
-    foreach (const MountInfoTable::Entry& entry, table.entries) {
-      if (entry.target == "/") {
-        CHECK_NONE(rootParentId);
-        rootParentId = entry.parent;
+      // It is legal to have a mount info table entry whose ID is the
+      // same as its parent ID. This can happen if, for example, the
+      // system boots from network and then keeps the original / in RAM.
+      // To avoid cycles when walking the mount hierarchy, we skip such
+      // entries in the children map.
+      if (ent.parent != ent.id) {
+        children[ent.parent].push_back(ent.id);
       }
-      parentToChildren[entry.parent].push_back(entry);
     }
 
-    // Walk the hashmap and construct a list of entries sorted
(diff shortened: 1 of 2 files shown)


src/linux/fs.cpp
```diff
@@ -193,54 +193,43 @@ Try<MountInfoTable> MountInfoTable::read(
...
-        // happen (for example), if a system boots from the network
-        // and then keeps the original `/` in RAM. To avoid cycles
-        // when walking the mount hierarchy, we only recurse into our
-        // children if this case is not satisfied.
-        if (parentId != entry.id) {
-          sortFrom(entry.id);
+    // Entries without a visible parent (e.g. in a chroot) or referring
+    // to themselves are roots of visible mount trees. Sort them in BFS
+    // order.
+    list<MountInfoTable::Entry> sorted;
```
bmahler, line 216: s/list/vector/

consider reserve()ing here since we know the end-size of the vector as well

src/linux/fs.cpp
```diff
@@ -193,54 +193,43 @@ Try<MountInfoTable> MountInfoTable::read(
...
-    // We know the node with a parent id of
-    // `rootParentId` is the root mount point.
-    CHECK_SOME(rootParentId);
-    sortFrom(rootParentId.get());
+    }
 
-    table.entries = std::move(sortedEntries);
+    CHECK_EQ(sorted.size(), table.entries.size())
+      << "Possible cycle in mount info table: " << std::endl << lines;
+    table.entries.assign(sorted.begin(), sorted.end());
```
bmahler, line 232: it's also a bit unfortunate that we have to copy the entries into the sorted vector, then copy it back into the table entries, perhaps we can std::move the sorted vector into table.entries to avoid at least the second copy?

```
table.entries = std::move(sorted);
```

src/linux/fs.cpp
```diff
@@ -193,54 +193,44 @@ Try<MountInfoTable> MountInfoTable::read(
...
-        // and then keeps the original `/` in RAM. To avoid cycles
-        // when walking the mount hierarchy, we only recurse into our
-        // children if this case is not satisfied.
-        if (parentId != entry.id) {
-          sortFrom(entry.id);
+    // Entries without a visible parent (e.g. in a chroot) or referring
+    // to themselves are roots of visible mount trees. Sort them in BFS
+    // order.
+    list<MountInfoTable::Entry> sorted;
+    sorted.reverse(table.entries.size());
```
bmahler, line 217: does this compile? s/reverse/reserve/ and s/list/vector/ on the previous line

bmahler: Just realized you probably were not compiling the network isolator (and therefore also not running tests), you need to pass an additional configure flag (and the tests require root after that as well):

```
--enable-network-ports-isolator
```

PR #469, opened 2024-01-16
Added LevelDB compaction after replicated log truncation.

DevinLeamy:
LevelDB compaction algorithm doesn't seem to work well with out key
usage pattern because the background compaction is not triggered [1].
Because of that the replicated log storage grows over time until it
consumes all disk space on the partition. As a workaround we can
manually invoke leveldb::DB::CompactRange() every time a truncation
action is persisted and some stored keys are removed. This workaround
can be turned off by setting the MESOS_LOG_AUTO_COMPACT_DISABLED=1 environment
variable.

[1] https://github.com/google/leveldb/issues/603

See MESOS-184 and https://github.com/google/leveldb/issues/603 for details.

### DIFF
--- a/src/log/leveldb.cpp
+++ b/src/log/leveldb.cpp
@@ -30,6 +30,8 @@
 #include <stout/strings.hpp>
 #include <stout/unreachable.hpp>
 
+#include <stout/os/getenv.hpp>
+
 #include "log/leveldb.hpp"
 
 using std::string;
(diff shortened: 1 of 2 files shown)


DevinLeamy:
Fixed commit message. @bmahler

bmahler:
I believe this was actually sent out for review but not landed, there should be 2 related commits: https://issues.apache.org/jira/browse/MESOS-184

Can you look through the feedback on them here:
https://reviews.apache.org/r/68089/
https://reviews.apache.org/r/68090/

DevinLeamy:
Updated commit message and environment variables.

src/log/leveldb.cpp
```diff
@@ -156,7 +158,10 @@ static string encode(uint64_t position, bool adjust = true)
 LevelDBStorage::LevelDBStorage()
   : db(nullptr), first(None())
 {
-  // Nothing to see here.
+  Option<string> disableAutoCompactVar = os::getenv("MESOS_LOG_AUTO_COMPACT_DISABLED");
+  if (autoCompactVar.isSome() && autoCompactVar.get() == "1") {
+    autoCompact = false;
+  }
```
bmahler, line 164: the names don't line up anymore? we can use getOrElse here too:

```suggestion
  // We provide an escape hatch in case some users experience issues
  // with auto-compaction. This is currently only available as an environment
  // variable since it's easier than exposing a flag up through the JNI layer.
  Option<string> disable = os::getenv("MESOS_LOG_AUTO_COMPACT_DISABLED");
  if (disable.getOrElse("0") == "1") {
    autoCompact = false;
  }
```

bmahler (changes requested):

PR #474, opened 2024-01-17
Added rates collector to network/port_mapping isolator.

DevinLeamy:
Added rates collector to network/port_mapping isolator.
Introduces RatesCollector class to network/port_mapping to (new)
RatePercentiles statistics on a variety of rates.

Statistics are derived from:
- https://docs.kernel.org/networking/statistics.html#c.rtnl_link_stats64

Queried through `rtnl_link_get_stat` in `src/linux/routing/link/link.cpp`.

### DIFF
--- a/include/mesos/mesos.proto
+++ b/include/mesos/mesos.proto
@@ -1888,6 +1888,54 @@ message ResourceStatistics {
   optional uint64 net_tx_burst_rate_limit = 47;
   optional uint64 net_tx_burst_size = 48;
 
+  message RatePercentiles {
+    optional uint64 min = 1;
+    optional uint64 max = 2;
+    optional uint64 p50 = 3;
+    optional uint64 p90 = 4;
+    optional uint64 p95 = 5;
+    optional uint64 p99 = 6;
+    optional uint64 p999 = 7;
+    optional uint64 p9999 = 8;
+    optional uint64 samples = 9;
+  }
+
+  // Network rate statistics measured in bytes per second
+  // or packets per second. 
+  // 
+  // Rates are sampled every {sampling_interval_secs}. A 
+  // time series is created out of the samples taken over
+  // a moving sampling window of {sampling_window_secs}. 
+  // Percentiles for each time series are exposed through 
+  // RatePercentiles.
+  //
+  // Linux documentation for more information:
+  // https://docs.kernel.org/networking/statistics.html#c.rtnl_link_stats64
+  message RateStatistics {
+    // Bytes received per second.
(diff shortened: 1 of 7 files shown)


include/mesos/mesos.proto
```diff
@@ -1888,6 +1888,33 @@ message ResourceStatistics {
...
+    optional RatePercentiles tx_packet_rate = 6;
+    optional RatePercentiles tx_drop_rate = 7;
+    optional RatePercentiles tx_error_rate = 8;
+    optional double sampling_window_secs = 9;
+    optional double sampling_interval_secs = 10;
+  }
+
+  optional RateStatistics net_rate_statistics = 51;
+
   // The kernel keeps track of RTT (round-trip time) for its TCP
```
DevinLeamy, line 1918: Wondering whether these new protobuf messages should be copied into `v1/mesos.proto` to maintain compatibility.

DevinLeamy: Copied over.

src/slave/containerizer/mesos/isolators/network/port_mapping.hpp
```diff
@@ -152,6 +153,59 @@ Result<routing::queueing::htb::cls::Config> recoverHTBConfig(
...
+  process::TimeSeries<uint64_t> rxErrors;
+
+  process::TimeSeries<uint64_t> txRates;
+  process::TimeSeries<uint64_t> txPackets;
+  process::TimeSeries<uint64_t> txDrops;
+  process::TimeSeries<uint64_t> txErrors;
+
+  // Previous sample and its timestamp for calculating rates.
+  Option<process::Time> previous;
+  hashmap<std::string, uint64_t> previousStatistics;
```
bmahler, line 202: Some follow-on optimizations: since in practice we're sampling every 50ms, it's probably worth avoiding the string keyed map here and instead use an enum indexed array like the library underneath us uses, which is significantly more efficient.

src/slave/containerizer/mesos/isolators/network/port_mapping.cpp
```diff
@@ -1431,6 +1431,279 @@ int PortMappingHTBConfig::execute()
...
+  }
+}
+
+
+void PercentileRatesCollector::sample(
+    const Time& ts, hashmap<string, uint64_t>&& statistics)
+{
+  if (previous.isSome() && previous.get() < ts) {
+    // We sample statistics on the host end of the veth pair, so we
+    // need to reverse RX and RX to get statistics inside the
```
bmahler, line 1520: RX and TX

PR #475, opened 2024-01-17
Auto-enable limit enforcement feature on the XFS file system, to enforce set quotas.

DevinLeamy:
If quotas are set for the XFS disk isolator they will only be
enforced if the limit enforcement feature is enabled, and it
is not always enabled by default. This change will auto-enable
quota enforcement on the XFS disk isolator, to get the expected
quota enforcement behavior.

### DIFF
--- a/src/slave/containerizer/mesos/isolators/xfs/disk.cpp
+++ b/src/slave/containerizer/mesos/isolators/xfs/disk.cpp
@@ -136,6 +136,12 @@ static Option<Bytes> getSandboxDisk(
 
 Try<Isolator*> XfsDiskIsolatorProcess::create(const Flags& flags)
 {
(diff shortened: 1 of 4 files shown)


src/slave/containerizer/mesos/isolators/xfs/disk.cpp
```diff
@@ -136,6 +136,12 @@ static Option<Bytes> getSandboxDisk(
 
 Try<Isolator*> XfsDiskIsolatorProcess::create(const Flags& flags)
 {
+  Try<Nothing> enable = xfs::enableQuota(flags.work_dir);
```
DevinLeamy, line 139: @bmahler Should there be a flag and/or environment variable to explicitly disable quota enforcement? Can't recall where we landed on this.

PR #477, opened 2024-01-22
Improved performance of RatesCollector in network/port_mapping isolator.

DevinLeamy:
To get a single link info utilities in routing::link pull all configured links infos from the kernel. With this change RatesCollector pulls the link cache only once per sampling round.

### DIFF
--- a/src/slave/containerizer/mesos/isolators/network/port_mapping.cpp
+++ b/src/slave/containerizer/mesos/isolators/network/port_mapping.cpp
@@ -62,6 +62,7 @@
 #include "linux/fs.hpp"
 #include "linux/ns.hpp"
 
+#include "linux/routing/internal.hpp"
 #include "linux/routing/route.hpp"
 #include "linux/routing/utils.hpp"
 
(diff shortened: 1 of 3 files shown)


src/slave/containerizer/mesos/isolators/network/port_mapping.cpp
```diff
@@ -1689,13 +1684,54 @@ class RatesCollectorProcess : public Process<RatesCollectorProcess>
...
 
 private:
-  void schedule()
+  // This method is mostly a copy of routing::link::statistics(), but
+  // with the ability to use the existing links cache. It was copied
+  // here to preserve libnl-agnostic interface of routing package and
+  // ease the cleanup when this collector will be replaced with eBPF.
+  static hashmap<string, uint64_t> statistics(
+      const Netlink<struct nl_cache>& cache,
+      const string& link)
```
DevinLeamy, line 1693: This can be removed in favor of using `this->link` directly. Any reason not to?

PR #483, opened 2024-01-23
Added auto calculation of egress_rate_per_cpu.

DevinLeamy:
Currently egress rate limit scaling requires manual configuration of per CPU rate limit. While it is more predictable and gives rounder values, in a heterogeneous environment where hosts with one NIC type have different CPUs it may be cumbersome to configure.

This commit allows you to pass an "auto" value to the --egress_rate_per_cpu flag of the isolator
```
--egress_rate_per_cpu=auto
```
to automatically calculate the value. The value is calculated by determining the link speed using the /sys/class/net/<iface>/speed interface (https://www.kernel.org/doc/Documentation/ABI/testing/sysfs-class-net) and dividing the speed by the number of CPUs resources that are available.  --egress_rate_per_cpu=auto will fail if there are 0 CPU resources.

The `network_link_speed` flag
```
--network_link_speed=<link speed in bytes/second>
```
can be used to override link speed detection when --egress_rate_per_cpu=auto.

### DIFF
--- a/src/slave/containerizer/mesos/isolators/network/port_mapping.cpp
+++ b/src/slave/containerizer/mesos/isolators/network/port_mapping.cpp
@@ -2246,6 +2246,50 @@ Try<Isolator*> PortMappingIsolatorProcess::create(const Flags& flags)
         " maximum ingress rate.");
   }
 
+  Option<Bytes> egressRatePerCpu;
+  if (flags.egress_rate_per_cpu.isSome()) {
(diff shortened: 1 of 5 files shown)


bmahler (changes requested):
Looks good, but can you update `--network_link_speed` flag help to mention this additional "auto" flag?

DevinLeamy:
> Looks good, but can you update `--network_link_speed` flag help to mention this additional "auto" flag?

Updated.

PR #485, opened 2024-01-23
Added per-CPU rate limit metrics.

DevinLeamy:
The new metrics show the amount of egress and ingress bandwidth that a container gets per whole unit of its CPU reservation. The values are 0 when limit scaling is not enabled.

Metric names:
```
port_mapping/per_cpu_egress_rate_limit
port_mapping/per_cpu_ingress_rate_limit
```


src/tests/containerizer/port_mapping_tests.cpp
```diff
@@ -2203,6 +2219,9 @@ TEST_F(PortMappingIsolatorTest, ROOT_ScaleIngressWithCPUAutoConfig)
   // Kill the container
   AWAIT_READY(launcher.get()->destroy(containerId1));
   AWAIT_READY(isolator.get()->cleanup(containerId1));
+
+  delete launcher.get();
+  delete isolator.get();
```
bmahler, line 2224: there's some extra deletes in here, can you split this out into its own commit or PR?

PR #486, opened 2024-01-23
[FIX] Rate overflow in port_mapping isolator.

DevinLeamy:
Commits:
- Fix 32-bit overflow in the port mapping isolator.
- [Compile time] Enforce use of libnl3 >= 3.5.
- [Run time] Enforce use of libnl3 >= 3.5.
- Make use of new libnl3 64 bit API to eliminate potential overflows.

The port mapping isolator currently uses uint32_t to store container rate limits as B/s. The use of this data type is dictated by the current libnl interface, that the isolator uses to interface with netlink.

Container preparation phase in the isolator involves setting up TX rate limiting HTB class using tc command that wants rate to be represented as bits/s. The conversion is performed by a simple multiplication of the rate by a literal 8. The result of such multiplication is uint32_t that can hold a maximum value of 4294967295, which is not enough to represent speeds available to us on modern 25G links as bits/s.

This change fixes the described overflow issue. However, representing network rates as 32bit integers still presents a problem on hosts with more advanced links, e.g. 100G. Hence, we update the required version of libnl to libnl >=3.5 to get access to the APIs uint64_t rates, eliminating potential overflows.

### DIFF
--- a/src/slave/containerizer/mesos/isolators/network/port_mapping.cpp
+++ b/src/slave/containerizer/mesos/isolators/network/port_mapping.cpp
@@ -5077,7 +5077,8 @@ Option<htb::cls::Config> PortMappingIsolatorProcess::egressHTBConfig(
   if (flags.egress_rate_limit_per_container.isSome()) {
     rate = flags.egress_rate_limit_per_container.get();
   } else if (egressRatePerCpu.isSome()) {
-    rate = egressRatePerCpu.get() * floor(resources.cpus().getOrElse(0));
+    rate = egressRatePerCpu.get() *
+      static_cast<uint64_t>(resources.cpus().getOrElse(0));
   } else {
     return None();
   }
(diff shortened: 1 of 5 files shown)


src/slave/containerizer/mesos/isolators/network/port_mapping.cpp
```diff
@@ -5292,10 +5294,11 @@ string PortMappingIsolatorProcess::scripts(Info* info)
     script << "tc class add dev " << eth0 << " parent "
            << CONTAINER_TX_HTB_HANDLE << " classid "
            << CONTAINER_TX_HTB_CLASS_ID << " htb rate "
-           << info->egressConfig->rate * 8 << "bit";
+           << static_cast<uint64_t>(info->egressConfig->rate) * 8 << "bit";
     if (info->egressConfig->ceil.isSome()) {
       script << " ceil "
-             << info->egressConfig->ceil.get() * 8 << "bit";
+             << static_cast<uint64_t>(info->egressConfig->ceil.get()) * 8
```
bmahler, line 5300: no longer need the static cast here when we've bumped up ceil to 64 bits?

src/slave/containerizer/mesos/isolators/network/port_mapping.cpp
```diff
@@ -5292,10 +5294,11 @@ string PortMappingIsolatorProcess::scripts(Info* info)
     script << "tc class add dev " << eth0 << " parent "
            << CONTAINER_TX_HTB_HANDLE << " classid "
            << CONTAINER_TX_HTB_CLASS_ID << " htb rate "
-           << info->egressConfig->rate * 8 << "bit";
+           << static_cast<uint64_t>(info->egressConfig->rate) * 8 << "bit";
```
bmahler, line 5297: no longer need the static cast here when we've bumped up rate to 64 bits?

bmahler (commented):
Overall looks good! Just noticed that we can take back out the static casts that become unnecessary in the later commit.

PR #487, opened 2024-01-23
Introduce support for HTB class and expose configuration.

DevinLeamy:
Added support for HTB ([Hierarchical Token Bucket](https://en.wikipedia.org/wiki/Token_bucket#Hierarchical_token_bucket)) packet scheduling by introducing an HTB queuing class and methods to update and access it's configuration.

The HTB implementation uses [libnl](https://www.infradead.org/~tgr/libnl/doc/api/group__qdisc__htb.html) and is used by the port mapping isolator.

### DIFF
--- a/src/linux/routing/queueing/class.hpp
+++ b/src/linux/routing/queueing/class.hpp
@@ -1,0 +1,61 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef __LINUX_ROUTING_QUEUEING_CLASS_HPP__
+#define __LINUX_ROUTING_QUEUEING_CLASS_HPP__
+
+#include <string>
+
+#include <netlink/route/class.h>
+
+#include <stout/option.hpp>
+
+#include "linux/routing/handle.hpp"
+#include "linux/routing/queueing/internal.hpp"
+
+namespace routing {
+
+template <>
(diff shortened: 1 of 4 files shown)


src/linux/routing/queueing/class.hpp
```diff
@@ -0,0 +1,61 @@
...
+#include "linux/routing/handle.hpp"
+#include "linux/routing/queueing/internal.hpp"
+
+namespace routing {
+
+template <>
+inline void cleanup(struct rtnl_class* cls)
+{
+  rtnl_class_put(cls);
+}
```
bmahler, line 35: just a note, this just seems to be the way to free the struct after an rtnl_class_alloc, it's a bit weird to see this function here without a corresponding alloc function, but maybe those come out from a different place

I would have expected maybe something like the Class struct below to alloc / free rtnl_class with a shared pointer around it to allow copying the Class struct. We don't need to change this in this patch just making some observations around the interface here.

src/linux/routing/queueing/internal.hpp
```diff
@@ -118,7 +120,69 @@ Try<Netlink<struct rtnl_qdisc>> encodeDiscipline(
...
+
+// Encodes a queueing class (in our representation) to a libnl
+// queueing class (rtnl_class). We use templating here so that it works
+// for any type of queueing class.
+template <typename Config>
+Try<Netlink<struct rtnl_class>> encode(
+    const Netlink<struct rtnl_link>& link,
+    const Class<Config>& config)
+{
+  struct rtnl_class* c = rtnl_class_alloc();
```
bmahler, line 152: here is that alloc I'm referring to that needs a corresponding put, but it looks like we make the caller deal with when to call the cleanup function

src/linux/routing/queueing/internal.hpp
```diff
@@ -118,7 +120,69 @@ Try<Netlink<struct rtnl_qdisc>> encodeDiscipline(
...
+    return Error(
+        "Failed to set the kind of the queueing class: " +
+        std::string(nl_geterror(error)));
+  }
+
+  // Perform queueuing class specific encoding.
+  Try<Nothing> encoding = encode(cls, config.config);
+  if (encoding.isError()) {
+    return Error("Failed to encode the queueing class: " + encoding.error());
+  }
```
bmahler, line 177: along the same line of thinking, aren't these error cases leaking the memory..? we've done the rtnl_class_alloc() but we're erroring out here and no one is going to free it?

bmahler (commented):
Let's follow up with an audit of the port mapping isolator code to check for memory leaks like I'm pointing out here, it's possible these don't happen in practice due to the error cases not being encountered, but we need to fix these leaks :)

PR #489, opened 2024-01-24
Enforce libnl3 >= 3.5 to prevent overflow errors in the port mapping isolator.

DevinLeamy:
libnl 3.5 changes the interface for link rates (bytes/s) from using a 32 bit integer to a 64 bit integer. Given that modern links can be on the order of 100 GB/s, using a 32 bit integer introduces a potential for overflow. This change prevents this kind of overflow from happening.

Link rates are queried through libnl in the port mapping isolator.

Relevant libnl commit:
- https://github.com/thom311/libnl/commit/4cf69a1b7b2518d18e1a62c79a461bca109132fc

### DIFF
--- a/m4/libnl3.m4
+++ b/m4/libnl3.m4
@@ -23,7 +23,7 @@ AC_DEFUN([MESOS_MSG_LIBNL3_ERROR], [
 
 AC_MSG_ERROR([$1
 -------------------------------------------------------------------
-Please install libnl3 (version 3.2.26 or higher):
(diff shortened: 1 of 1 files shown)


bmahler (changes requested):
Probably better to do this as a pre-processor based check, similar to how you see protobuf perform checks in any generated .pb.h header. This catches autotools and cmake builds and reduces build complexity which is nice, since it's purely in our code files.

DevinLeamy:
Closed in favor of https://github.com/apache/mesos/pull/486.

PR #492, opened 2024-01-26
Added ENABLE_JEMALLOC_ALLOCATOR macro.

DevinLeamy:
Summary:
The macro is used in the agent code to determine if jemalloc was enabled in the build and provide a custom allocator configuration in the form of a global malloc_conf variable.

This commit follows from https://github.com/apache/mesos/commit/2836057c8881a62139b5db1f175d3d027bf2e645. View the commit for more details.

Test Plan:
Modified the agent code to print whether the macro is defined or not and tested on an integration test host.

### DIFF
--- a/cmake/CompilationConfigure.cmake
+++ b/cmake/CompilationConfigure.cmake
@@ -491,6 +491,10 @@ if (LINUX)
     "Use jemalloc as memory allocator for the master and agent binaries."
     FALSE)
 
+  if (ENABLE_JEMALLOC_ALLOCATOR)
+    add_definitions(-DENABLE_JEMALLOC_ALLOCATOR)
+  endif ()
+
   option(ENABLE_XFS_DISK_ISOLATOR
     "Whether to enable the XFS disk isolator."
     FALSE)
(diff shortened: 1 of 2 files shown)


bmahler:
Can you update this to handle the cmake build as well?

cmake/CompilationConfigure.cmake
```diff
@@ -509,6 +509,10 @@ if (LINUX)
     "Whether to enable the port mapping isolator."
     FALSE)
 
+  if (ENABLE_JEMALLOC_ALLOCATOR)
+    add_definitions(-DENABLE_JEMALLOC_ALLOCATOR)
```
DevinLeamy, line 513: As per: https://cmake.org/cmake/help/v3.8/command/add_definitions.html?highlight=i

DevinLeamy:
> Can you update this to handle the cmake build as well?

Updated :+1:

cmake/CompilationConfigure.cmake
bmahler: Thanks! Any reason you don't have it further up next to the option definition? Similar to how xfs's option and if condition are next to each other?

PR #494, opened 2024-02-14
[cgroups2] Bootstrapping cgroups v2

DevinLeamy:
### Bootstrapping for cgroups2
- `linux/cgroups2.hpp`
- `linux/cgroups2.cpp`
- `tests/containerizer/cgroups2_tests.cpp`

### Overview

The main namespace for cgroups v2 is `cgroups2`. The top-level control are in `cgroups2::control::<name>`. Controls for subsystems will similarly be put in `cgroups2::<subsystem>::control::<name>`. The patch adds the ability to mount, unmount, query subsystems, enable subsystems, and introduces utilities for making these operations easier. Tests are also introduced. 

---

Depends on: https://github.com/apache/mesos/pull/493


qianzhangxa:
@DevinLeamy It seems what this PR does is same as the patch chain that you submitted to the review board (starting from https://reviews.apache.org/r/74869/), so which one should we review?

DevinLeamy:
> @DevinLeamy It seems what this PR does is same as the patch chain that you submitted to the review board (starting from https://reviews.apache.org/r/74869/), so which one should we review?

 Hi @qianzhangxa, you're right, they are the same patches. I was deliberating between using GitHub and Review Board. I'll be using Review Board, for the time being, so I'll close out this PR :+1:

PR #501, opened 2024-03-07
[ebpf] Introduced API for loading eBPF programs into the kernel.

DevinLeamy:
Introduces the `ebpf::ProgramBuilder` and `ebpf::Program` classes, and `ebpf::load(program)` function, allowing generic eBPF programs to be loaded into the kernel.

### DIFF
--- a/src/linux/ebpf.cpp
+++ b/src/linux/ebpf.cpp
@@ -15,3 +15,85 @@
 // limitations under the License.
 
 #include "linux/ebpf.hpp"
+
+#include <linux/bpf.h>
+#include <sys/syscall.h>
+
+#include <string>
+#include <vector>
+
+#include "stout/check.hpp"
+#include "stout/error.hpp"
+#include "stout/try.hpp"
+
+using std::string;
+using std::vector;
+
+namespace ebpf {
+
+Try<int, ErrnoError> bpf(int cmd, bpf_attr* attr, size_t size)
+{
+  // We retry the system call `attempts` times on EAGAIN. The default is 5,
+  // as per what's done by libbpf:
+  // https://github.com/libbpf/libbpf/blob/master/src/bpf.h#L71-L75
+  int result, attempts = 5;
+  do {
+    // glibc does not expose the bpf() function, requiring us to make the
+    // syscall directly: https://lwn.net/Articles/655028/
+    result = (int)syscall(__NR_bpf, cmd, attr, size);
+  } while (result == -1 && errno == EAGAIN && --attempts > 0);
+
+  if (result == -1) {
+    return ErrnoError();
+  }
+  return result;
+}
+
+
+Program::Program(bpf_prog_type _type) : type(_type) {}
+
+
+void Program::append(vector<bpf_insn>&& instructions)
+{
+  program.insert(
+      program.end(),
+      std::make_move_iterator(instructions.begin()),
+      std::make_move_iterator(instructions.end()));
+}
+
+
+Try<int> load(const Program& program)
+{
+  bpf_attr attribute;
+  std::memset(&attribute, 0, sizeof(attribute));
+  attribute.prog_type = program.type;
+  attribute.insn_cnt = program.program.size();
+  attribute.insns = reinterpret_cast<uint64_t>(program.program.data());
+  attribute.license = reinterpret_cast<uint64_t>("Apache 2.0");
+
+  Try<int, ErrnoError> fd = bpf(BPF_PROG_LOAD, &attribute, sizeof(attribute));
+
+  if (fd.isError() && fd.error().code == EACCES) {
+    // If bpf() fails with EACCES (a verifier error) the system call is called
+    // again with an additional buffer to capture the verifier error logs.
+    string verifier_logs(8196, '\0');
+    attribute.log_level = 1;
+    attribute.log_buf = reinterpret_cast<uint64_t>(&verifier_logs.front());
+    attribute.log_size = verifier_logs.size();
+
+    fd = bpf(BPF_PROG_LOAD, &attribute, sizeof(attribute));
+
+    CHECK_ERROR(fd);
+    CHECK_EQ(EACCES, fd.error().code)
+      << "Expected BPF syscall to fail again with EACCES";
+    return Error(string("BPF verifier failed: ") + verifier_logs.c_str());
+  }
+
+  if (fd.isError()) {
+    return Error("Unexpected error from BPF syscall: " + fd.error().message);
+  }
+
+  return *fd;
+}
+
+} // namespace ebpf {
--- a/src/linux/ebpf.hpp
+++ b/src/linux/ebpf.hpp
@@ -14,7 +14,123 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+// TODO(dleamy): Look into using libbpf: https://github.com/libbpf/libbpf
+//               to simplify and or replace the low-level BPF operations.
+
 #ifndef __EBPF_HPP__
 #define __EBPF_HPP__
+
+#include <linux/bpf.h>
+
+#include <vector>
+
+#include "stout/try.hpp"
+
+namespace ebpf {
+
+// eBPF program.
+class Program
+{
+public:
+  explicit Program(bpf_prog_type type);
+
+  // Append instructions to the end of the eBPF program.
+  void append(std::vector<bpf_insn>&& instructions);
+
+  // Type of eBPF program.
+  const bpf_prog_type type;
+
+  // Instructions of the eBPF program.
+  std::vector<bpf_insn> program;
+};
+
+
+// Loads the provided eBPF program into the kernel and returns the file
+// descriptor of loaded program.
+Try<int> load(const Program& program);
+
+
+// Utility macros for constructing eBPF instructions.
+#define BPF_ALU32_IMM(OP, DST, IMM)       \
+  ((bpf_insn){                            \
+    .code = BPF_ALU | BPF_OP(OP) | BPF_K, \
+    .dst_reg = DST,                       \
+    .src_reg = 0,                         \
+    .off = 0,                             \
+    .imm = IMM})
+
+
+#define BPF_LDX_MEM(SIZE, DST, SRC, OFF)        \
+  ((bpf_insn){                                  \
+    .code = BPF_LDX | BPF_SIZE(SIZE) | BPF_MEM, \
+    .dst_reg = DST,                             \
+    .src_reg = SRC,                             \
+    .off = OFF,                                 \
+    .imm = 0})
+
+
+#define BPF_MOV64_REG(DST, SRC)          \
+  ((bpf_insn){                           \
+    .code = BPF_ALU64 | BPF_MOV | BPF_X, \
+    .dst_reg = DST,                      \
+    .src_reg = SRC,                      \
+    .off = 0,                            \
+    .imm = 0})
+
+
+#define BPF_JMP_A(OFF)        \
+  ((bpf_insn){                \
+    .code = BPF_JMP | BPF_JA, \
+    .dst_reg = 0,             \
+    .src_reg = 0,             \
+    .off = OFF,               \
+    .imm = 0})
+
+
+#define BPF_JMP_IMM(OP, DST, IMM, OFF)    \
+  ((bpf_insn){                            \
+    .code = BPF_JMP | BPF_OP(OP) | BPF_K, \
+    .dst_reg = DST,                       \
+    .src_reg = 0,                         \
+    .off = OFF,                           \
+    .imm = IMM})
+
+
+#define BPF_JMP_REG(OP, DST, SRC, OFF)    \
+  ((bpf_insn){                            \
+    .code = BPF_JMP | BPF_OP(OP) | BPF_X, \
+    .dst_reg = DST,                       \
+    .src_reg = SRC,                       \
+    .off = OFF,                           \
+    .imm = 0})
+
+
+#define BPF_MOV64_IMM(DST, IMM)          \
+  ((bpf_insn){                           \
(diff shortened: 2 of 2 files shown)


src/linux/ebpf.cpp
```diff
@@ -14,4 +14,99 @@
...
+{
+  union bpf_attr attribute = program.attribute;
+  Try<int> fd = 0; // placeholder because Try<int>() DNE
+  string logs = "";
+
+  if (logErrors) {
+    const int LOG_SIZE = 8196;
+    const int LOG_LEVEL = 1;
+
+    char* log_buffer = (char*)malloc(LOG_SIZE);
```
bmahler, line 95: this is a memory leak! we could stack (or unique_ptr + new) allocate it inside the if EACCESS if condition:

```cpp
char verifier_log_buffer[VERIFIER_LOG_SIZE];

// Ensure the log is null terminated if the verifier doesn't write anything.
verifier_log_buffer[0] = '\0';

attribute.log_buf = static_cast<uint64_t>(verifier_log_buffer);

// We don't bother looking for ENOSPC because we use a very large buffer size.
XXX

// Maybe check that it's EACCESS again..?
ErrnoError error;
CHECK_EQ(EACCESS, error.errno) << "Expected verifier error again, got: " << error;

return Error("bpf verifier rejected the program: " + string(verifier_log_buffer))
```

src/linux/ebpf.cpp
```diff
@@ -14,4 +14,99 @@
...
+
+
+Try<int> load(const Program& program, bool logErrors)
+{
+  union bpf_attr attribute = program.attribute;
+  Try<int> fd = 0; // placeholder because Try<int>() DNE
+  string logs = "";
+
+  if (logErrors) {
+    const int LOG_SIZE = 8196;
```
bmahler, line 92: if we use a big size like this, consider using a vector<char> here as the buffer so it's on the heap and managed, and zeroed for us

bmahler (changes requested):

src/linux/ebpf.cpp
```diff
@@ -14,4 +14,99 @@
...
+      "BFP syscall failed with error '" + string(std::strerror(errno)) + "'");
+  }
+  return result;
+}
+
+
+uint64_t pointer_to_int(const void* pointer)
+{
+  return (uint64_t)(unsigned long)pointer;
+}
```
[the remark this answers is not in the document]
DevinLeamy: This pointer cast `static_cast<uint64_t>(program.data())` doesn't work. However, pointers can directly converted into integer types and back during `reinterpret_cast` (https://stackoverflow.com/a/21250110).

```cpp
reinterpret_cast<uint64_t>(program.data())
```

src/linux/ebpf.cpp
```diff
@@ -14,4 +14,99 @@
...
+
+namespace internal {
+
+// Wrapper around the `bpf` syscall.
+Try<int> syscall_bpf(int cmd, union bpf_attr* attr, size_t size)
+{
+  int result = syscall(__NR_bpf, cmd, attr, size);
+  if (result == -1) {
+    return Error(
+      "BFP syscall failed with error '" + string(std::strerror(errno)) + "'");
```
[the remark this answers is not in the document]
DevinLeamy: Kept the message but used `ErrnoError` to get the error string.

src/linux/ebpf.cpp
```diff
@@ -14,4 +14,99 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#include "linux/ebpf.hpp"
\ No newline at end of file
+#include "linux/ebpf.hpp"
+
+#include <errno.h>
+#include <linux/bpf.h>
+#include <sys/syscall.h>
```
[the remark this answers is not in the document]
DevinLeamy: Kept because the direct syscall is not exposed through `<linux/bpf.h>`.

src/linux/ebpf.hpp
```diff
@@ -17,4 +17,136 @@
...
+  explicit ProgramBuilder(enum bpf_prog_type type);
+
+  // Append instructions to the end of the eBPF program.
+  void append(std::vector<struct bpf_insn> instructions);
+
+  // Number of instructions in the eBPF program.
+  size_t instructions() const;
+
+  // Build a loadable eBPF program.
+  Program build();
```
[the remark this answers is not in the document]
DevinLeamy: I opted to copy the program instructions from ProgramBuilder into Program on build, and then free the instructions in the Program's destructor. 

I saw having the `Builder` as useful for implementing the DeviceController where there's instructions we want to add to the program instructions before it becomes a `Program`. By having this separation we can do those additions in the `build()` step, so the caller doesn't have to remember to do some `finalize()` step before "building" (AKA creating the bpf_attr). The idea is make sure `Program`s are _truly_ loadable eBPF programs.

If that's not clear, here's the DeviceProgram I'm preparing for the device controller (note - the API is slightly different but it should make what I'm going for more evident): https://github.com/DevinLeamy/mesos/commit/50e4c0ff7cf806ae02e8dc03488b70d00412583f

src/linux/ebpf.cpp
```diff
@@ -14,4 +14,99 @@
...
+    attribute.log_size = LOG_SIZE;
+    fd = internal::syscall_bpf(BPF_PROG_LOAD, &attribute, sizeof(attribute));
+    logs = ": " + string(log_buffer);
+  } else {
+    fd = internal::syscall_bpf(BPF_PROG_LOAD, &attribute, sizeof(attribute));
+  }
+
+  if (fd.isError()) {
+    return Error("BPF syscall failed" + logs + ": " + fd.error());
+  }
```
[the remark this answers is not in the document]
DevinLeamy: Updated to work similarly to how you described. We attempt and then retry with logs if there's a validator error.

src/linux/ebpf.hpp
```diff
@@ -17,4 +17,136 @@
...
 #define __EBPF_HPP__
 
+#include <linux/bpf.h>
+
+#include <cstddef>
+#include <vector>
+
+#include "stout/nothing.hpp"
+#include "stout/try.hpp"
+
```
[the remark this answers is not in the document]
DevinLeamy: Added.

src/linux/ebpf.cpp
```diff
@@ -14,4 +14,126 @@
...
+  bpf_attr attribute;
+  std::memset(&attribute, 0, sizeof(attribute));
+
+  attribute.prog_type = type;
+  attribute.insn_cnt = program.size();
+  attribute.insns = reinterpret_cast<uint64_t>(program.data());
+  attribute.license = reinterpret_cast<uint64_t>("Apache 2.0");
+
+  return attribute;
+}
```
bmahler, line 70: this is still a dangerous function because the caller *must* keep Program alive longer than the bpf_attr returned by this function, and that's not obvious (e.g. this is called create and returns a value, rather than, say, a const ref) and easy to get wrong and cause a use-after-free bug

we could instead just expose the instructions and the license, and have `load(const Program& program)` build the bpf_attr struct needed for loading, this also avoids the need to do any moving since Program clearly outlives load

src/linux/ebpf.cpp
```diff
@@ -14,4 +14,126 @@
...
+
+bpf_attr Program::createAttribute() const
+{
+  bpf_attr attribute;
+  std::memset(&attribute, 0, sizeof(attribute));
+
+  attribute.prog_type = type;
+  attribute.insn_cnt = program.size();
+  attribute.insns = reinterpret_cast<uint64_t>(program.data());
+  attribute.license = reinterpret_cast<uint64_t>("Apache 2.0");
```
bmahler, line 67: AFAICT this is safe because C++ guarantees that string literals have static storage and are alive for the lifetime of the program, but it would be more obviously safe here to just have a const LICENSE string stored statically in this .cpp file.

src/linux/ebpf.cpp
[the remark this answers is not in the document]
bmahler: yeah, I think it's actually supposed to be glibc that exposes it per the lwn article, so we can reference that article below where we make a syscall

src/linux/ebpf.cpp
```diff
@@ -14,4 +14,126 @@
...
+#include "stout/try.hpp"
+
+using std::string;
+using std::vector;
+
+namespace ebpf {
+
+#define syscall_bpf(cmd, attr, size) (int) syscall(__NR_bpf, cmd, attr, size);
+
+// Wrapper around the BPF syscall.
```
bmahler, line 35: this is clear, no need for this comment

src/linux/ebpf.cpp
```diff
@@ -14,4 +14,126 @@
...
+Try<int> bpf(int cmd, bpf_attr* attr, size_t size)
+{
+  int result = syscall_bpf(cmd, attr, size);
+  if (result == -1) {
+    ErrnoError error;
+    return Error("BPF syscall failed with error '" + error.message + "'");
```
bmahler, line 41: per offline discussion, it would be good to deal with EAGAIN here, and reference where libbpf uses a default of 5 EAGAIN retries to show why we chose this number of retries:

https://github.com/libbpf/libbpf/blob/master/src/bpf.h#L71-L75

src/linux/ebpf.cpp
```diff
@@ -14,4 +14,126 @@
...
+  attribute.license = reinterpret_cast<uint64_t>("Apache 2.0");
+
+  return attribute;
+}
+
+
+Try<int> load(const Program&& program)
+{
+  bpf_attr attribute = program.createAttribute();
+  string logs = "";
```
bmahler, line 76: we shouldn't need this extra string and can use the verifier log buffer directly

src/linux/ebpf.cpp
```diff
@@ -14,4 +14,126 @@
...
+  // capture the verifier error logs.
+  //
+  // We attempt a maximum of 3 times, to avoid an infinite loop if 'EAGAIN' is
+  // returned multiple times.
+  int attempts = 0;
+  bool retry = false;
+  bool retriedWithLogs = false;
+
+  Try<int> fd = 0; // '0' because Try<int> has no default constructor.
+  do {
```
bmahler, line 91: once EAGAIN is moved into the wrapper, we can avoid a loop and the three loop related variables

src/linux/ebpf.cpp
```diff
@@ -14,4 +14,126 @@
...
+  // verification error - when we retry once with an additional buffer to
+  // capture the verifier error logs.
+  //
+  // We attempt a maximum of 3 times, to avoid an infinite loop if 'EAGAIN' is
+  // returned multiple times.
+  int attempts = 0;
+  bool retry = false;
+  bool retriedWithLogs = false;
+
+  Try<int> fd = 0; // '0' because Try<int> has no default constructor.
```
bmahler, line 90: once the loop is removed, this = 0 can be avoided by assigning it on the first of the two bpf calls

src/linux/ebpf.cpp
```diff
@@ -14,4 +14,126 @@
...
+        attribute.log_size = verifier_log_buffer.size();
+        fd = bpf(BPF_PROG_LOAD, &attribute, sizeof(attribute));
+        logs = ": " + string(verifier_log_buffer.data());
+        retry = true;
+        break;
+      }
+      default: {
+        logs = ": Unexpcted error from BPF syscall: " + error.message;
+        retry = false;
+        break;
```
bmahler, line 127: we won't need this anymore, we can just do a:

```
return Error("bpf syscall failed: " + fd.error());
...
return Error("bpf verifier failed: " + &verifier_log_buffer[0]);
```

src/linux/ebpf.cpp
```diff
@@ -14,4 +14,126 @@
...
+          break;
+        }
+
+        retry = true;
+        vector<char> verifier_log_buffer(8196);
+        attribute.log_level = 1;
+        attribute.log_buf =
+          reinterpret_cast<uint64_t>(verifier_log_buffer.data());
+        attribute.log_size = verifier_log_buffer.size();
+        fd = bpf(BPF_PROG_LOAD, &attribute, sizeof(attribute));
```
bmahler, line 119: perhaps add a CHECK here which requires using `Try<int, ErrnoError>` as the bpf return type:

```
CHECK_ERROR(fd);
CHECK_EQ(EACCESS, fd.error().code);
```

bmahler (changes requested):

src/linux/ebpf.cpp
DevinLeamy: Updated.

src/linux/ebpf.cpp
```diff
@@ -14,4 +14,90 @@
...
+
+Try<int, ErrnoError> bpf(int cmd, bpf_attr* attr, size_t size)
+{
+  // We retry the system call `attempts` times. The default is 5, as per
+  // what's done by libbpf:
+  // https://github.com/libbpf/libbpf/blob/master/src/bpf.h#L71-L75
+  int result, attempts = 5;
+  do {
+    result = (int)syscall(__NR_bpf, cmd, attr, size);
+  } while (result == EAGAIN && --attempts > 0);
```
bmahler, line 42: this needs to check errno

src/linux/ebpf.cpp
```diff
@@ -14,4 +14,90 @@
...
+
+Try<int, ErrnoError> bpf(int cmd, bpf_attr* attr, size_t size)
+{
+  // We retry the system call `attempts` times. The default is 5, as per
+  // what's done by libbpf:
+  // https://github.com/libbpf/libbpf/blob/master/src/bpf.h#L71-L75
+  int result, attempts = 5;
+  do {
+    result = (int)syscall(__NR_bpf, cmd, attr, size);
+  } while (errno == EAGAIN && --attempts > 0);
```
bmahler, line 42: this needs to check that result == -1 to check errno

bmahler (approved):

PR #503, opened 2024-03-08
[cgroups2] Add utility to list all the available subsystems in a cgroup.

DevinLeamy:
Interface to the "cgroup.controllers" control file.

### DIFF
--- a/src/linux/cgroups2.cpp
+++ b/src/linux/cgroups2.cpp
@@ -72,6 +72,27 @@ const std::string TYPE = "cgroup.type";
 
 } // namespace control {
 
+namespace controllers {
+
+// Find the available controllers (AKA subsystems) in the provided cgroup.
+Try<set<string>> available(const string& cgroup)
+{
+  Try<string> subsystems =
+    cgroups2::read(cgroup, cgroups2::control::CONTROLLERS);
+  if (subsystems.isError()) {
+    return Error(
+      "Failed to read cgroup.controllers in '" + cgroup +
+      "': " + subsystems.error());
+  }
+  vector<string> _subsystems = strings::split(subsystems.get(), " ");
+  set<string> systems(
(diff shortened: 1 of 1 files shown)


src/linux/cgroups2.cpp
```diff
@@ -72,6 +72,27 @@ const std::string TYPE = "cgroup.type";
 
 } // namespace control {
 
+namespace controllers {
+
+// Find the available controllers (AKA subsystems) in the provided cgroup.
+Try<set<string>> available(const string& cgroup)
+{
+  Try<string> subsystems =
```
bmahler, line 80: s/subsystems/read/

src/linux/cgroups2.cpp
```diff
@@ -72,6 +72,27 @@ const std::string TYPE = "cgroup.type";
...
+    cgroups2::read(cgroup, cgroups2::control::CONTROLLERS);
+  if (subsystems.isError()) {
+    return Error(
+      "Failed to read cgroup.controllers in '" + cgroup +
+      "': " + subsystems.error());
+  }
+  vector<string> _subsystems = strings::split(subsystems.get(), " ");
```
bmahler, line 87: can use * rather than .get() here

s/_//

src/linux/cgroups2.cpp
```diff
@@ -72,6 +72,27 @@ const std::string TYPE = "cgroup.type";
...
+  set<string> systems(
+    std::make_move_iterator(_subsystems.begin()),
+    std::make_move_iterator(_subsystems.end()));
+  return systems;
```
bmahler, line 91: rather than declaring the variable and immediately returning it, we can return it directly

bmahler (approved):
There are a couple of style edits needed as well due to the lack of completeness in clang-format but I'll make all these edits and land this for you shortly.

PR #505, opened 2024-03-08
[ebpf][cgroups2] Attaching a eBFP programs to a cgroups.

DevinLeamy:
Introduces `ebpf::cgroups2::attach()` which allows an `ebpf::Program()` to be attached to a cgroup, given the absolute path to the cgroup and the file descriptor of the loaded program.

This will be used by cgroups v2 Device Controller.

### DIFF
--- a/src/linux/ebpf.cpp
+++ b/src/linux/ebpf.cpp
@@ -24,6 +24,8 @@
 
 #include "stout/check.hpp"
 #include "stout/error.hpp"
+#include "stout/nothing.hpp"
+#include "stout/os.hpp"
 #include "stout/try.hpp"
 
 using std::string;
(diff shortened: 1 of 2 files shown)


src/linux/ebpf.cpp
```diff
@@ -96,4 +98,51 @@ Try<int> load(const Program& program)
...
 }
 
+namespace cgroups2 {
+
+Try<Nothing> attach(int fd, const string& cgroup)
+{
+  Try<int> cgroupFd = os::open(cgroup, O_DIRECTORY | O_RDONLY);
+  if (cgroupFd.isError()) {
+    return Error(
+        "Failed to find the file descriptor of '" + cgroup +
```
bmahler, line 108: Failed to open '...': ...

src/linux/ebpf.cpp
```diff
@@ -96,4 +98,51 @@ Try<int> load(const Program& program)
   return *fd;
 }
 
+namespace cgroups2 {
+
+Try<Nothing> attach(int fd, const string& cgroup)
+{
+  Try<int> cgroupFd = os::open(cgroup, O_DIRECTORY | O_RDONLY);
```
bmahler, line 105: we have to close this file descriptor always on exit of this function, or we will leak open file descriptors

src/linux/ebpf.cpp
```diff
@@ -96,4 +98,51 @@ Try<int> load(const Program& program)
...
+  //
+  // Order: oldest, ..., newest
+  // cgroup1: A, B
+  //   cgroup2: C
+  //   cgroup3: D, E
+  //     cgroup4: F
+  // cgroup4 run order: F, D, E, A, B
+  // cgroup2 run order: C, A, B
+  //
+  // src: https://elixir.bootlin.com/linux/latest/source/include/uapi/linux/bpf.h#L1090
```
bmahler, line 134: let's use a permalink here instead, or it will drift as the linux source evolves

src/linux/ebpf.cpp
```diff
@@ -96,4 +98,51 @@ Try<int> load(const Program& program)
...
+        "': " + cgroupFd.error());
+  }
+
+  bpf_attr attr;
+  memset(&attr, 0, sizeof(attr));
+  attr.attach_type = BPF_CGROUP_DEVICE;
+  attr.target_fd = *cgroupFd;
+  attr.attach_bpf_fd = fd;
+  // BPF_F_ALLOW_MULTI allows multiple eBPF programs to be attached to a single
+  // cgroup and determines the order in which attached programs will run.
```
bmahler, line 118: we don't want multiple programs within a single cgroup, rather we want to replace them, we can add a TODO to perform replacement for now

bmahler (approved):
I'll make the changes mentioned here and land this shortly.

PR #510, opened 2024-03-11
[cgroups2] Adds cgroups2::controllers::enabled() to obtain the set of enabled controllers.

DevinLeamy:
```cpp
// Get all the controllers that are enabled for a cgroup.
Try<std::set<std::string>> enabled(const std::string& cgroup);
```


bmahler (commented):
summary is a bit off? s/cgroups2::enabled()/cgroups2::controllers::enabled()/ and s/subsystems/controllers/

PR #512, opened 2024-03-11
[ebpf][cgroups2] Update ebpf::cgroups2::attach() to load and attach.

DevinLeamy:
Correctly handling eBPF file descriptors requires some subtle knowledge about how loading and unloading eBPF programs works.

Specifically, to unload an eBPF program there cannot be any open file descriptors to the program and the program cannot be attached.

To avoid requiring users of the API to know that the file descriptor returned by `ebpf::load()` needs to be closed before the loaded program can be detached, we remove `ebpf::load()` from the public interface and update `ebpf::cgroups2::load()` to:
  1. Load a program,
  2. attach the program, and
  3. close the loaded program's file descriptor, so detaching it will unload it.

### DIFF
--- a/src/linux/ebpf.cpp
+++ b/src/linux/ebpf.cpp
@@ -65,6 +65,8 @@ void Program::append(vector<bpf_insn>&& instructions)
 }
 
 
+// Load an eBPF program into the kernel and return the file
+// descriptor of the loaded program.
 Try<int> load(const Program& program)
 {
   bpf_attr attribute;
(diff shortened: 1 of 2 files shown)


src/linux/ebpf.cpp
```diff
@@ -65,6 +65,8 @@ void Program::append(vector<bpf_insn>&& instructions)
 }
 
 
+// Load an eBPF program into the kernel and return the file
+// descriptor of the loaded program.
```
bmahler, line 69: This is still left in the header?

src/linux/ebpf.cpp
```diff
@@ -152,6 +159,24 @@ Try<Nothing> attach(int fd, const string& cgroup)
...
+Try<Nothing> attach(const string& cgroup, const Program& program)
+{
+  Try<int> program_fd = ebpf::load(program);
+  if (program_fd.isError()) {
+    return Error("Failed to load eBPF program: " + program_fd.error());
+  }
+
+  Try<Nothing> _attach = attach(cgroup, *program_fd);
+  if (_attach.isError()) {
+    return Error("Failed to attach eBPF program: " + _attach.error());
```
bmahler, line 171: we're leaking the program fd in this branch

bmahler (commented):
Will make the edits and land it.

PR #514, opened 2024-03-12
[cgroups2][ebpf] Addresses formatting inconsistences.

DevinLeamy:
- Two free lines between functions.
- One free line between ending namespaces.
- One free line between methods.
- Ending namespace having format `} // namespace <name> {`.

### DIFF
--- a/src/linux/cgroups2.cpp
+++ b/src/linux/cgroups2.cpp
@@ -102,6 +102,7 @@ struct State
   }
 
   set<string> enabled()  const { return _enabled; }
+
   set<string> disabled() const { return _disabled; }
 
   bool enabled(const string& controller) const
@@ -279,4 +280,5 @@ Try<Nothing> enable(const string& cgroup, const vector<string>& controllers)
 }
 
 } // namespace controllers {
+
 } // namespace cgroups2 {
--- a/src/linux/cgroups2.hpp
+++ b/src/linux/cgroups2.hpp
@@ -33,13 +33,16 @@ const std::string ROOT_CGROUP = "";
 // Checks if cgroups2 is available on the system.
 bool enabled();
 
+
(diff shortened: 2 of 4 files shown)


bmahler:
we don't enforce all of these rules, but I'll get this landed with most of the changes:

* Two free lines between functions: I think this rule was primarily for .cpp files, but this is ok in hpp files too, we're probably inconsistent in the header files, for example I see that cgroups.hpp groups some related things with a single line, and then uses two lines in other places, it's pretty inconsistent.
* One free line between ending namespaces: is this a clang format rule? I'm fine with enforcing this, but I imagine we're not consistent about it.
* One free line between methods: not a hard rule, typically used for one liners, e.g. https://github.com/apache/mesos/blob/1.11.0/3rdparty/stout/include/stout/option.hpp#L116-L117
* Ending namespace having format } // namespace <name> {.: looks good

DevinLeamy:
> we don't enforce all of these rules, but I'll get this landed with most of the changes:
> 
> * Two free lines between functions: I think this rule was primarily for .cpp files, but this is ok in hpp files too, we're probably inconsistent in the header files, for example I see that cgroups.hpp groups some related things with a single line, and then uses two lines in other places, it's pretty inconsistent.
> * One free line between ending namespaces: is this a clang format rule? I'm fine with enforcing this, but I imagine we're not consistent about it.
> * One free line between methods: not a hard rule, typically used for one liners, e.g. https://github.com/apache/mesos/blob/1.11.0/3rdparty/stout/include/stout/option.hpp#L116-L117
> * Ending namespace having format } // namespace  {.: looks good

I based most of this on what I saw being done in `cgroups.hpp` and `cgroups.cpp`, which I think you might've noticed.

PR #518, opened 2024-03-15
Address clang-tidy errors.

DevinLeamy:
Fixes clang-tidy errors in ebpf and cgroups2.


src/linux/ebpf.cpp
```diff
@@ -91,7 +91,10 @@ Try<int> load(const Program& program)
     CHECK_ERROR(fd);
     CHECK_EQ(EACCES, fd.error().code)
       << "Expected BPF syscall to fail again with EACCES";
-    return Error(string("BPF verifier failed: ") + verifier_logs.c_str());
```
bmahler, line 94: can you keep the comment but use .data() instead? that seems a bit more appropriate here and I suspect won't produce the warning?

DevinLeamy: Updated.

PR #519, opened 2024-03-18
[cgroups2][ebpf] Configure device access permissions.

DevinLeamy:
Cgroups v2 uses eBPF to manage device access permissions.  

Device access is configured per-cgroup.

For each cgroup, we attach a `BPF_PROG_TYPE_CGROUP_DEVICE` program type to the `BPF_CGROUP_DEVICE` eBPF kernel hook. The attached program will be run for every device access made from within the cgroup. If the program returns a `0` exit code the device access is denied, otherwise it is allowed.

For more information about the Device Controller in cgroups v2: https://docs.kernel.org/admin-guide/cgroup-v2.html#device-controller

### DIFF
--- a/src/linux/cgroups2.cpp
+++ b/src/linux/cgroups2.cpp
@@ -25,9 +25,11 @@
 #include <stout/numify.hpp>
 #include <stout/os.hpp>
 #include <stout/path.hpp>
+#include <stout/unreachable.hpp>
 #include <stout/stringify.hpp>
 #include <stout/try.hpp>
 
+#include "linux/ebpf.hpp"
 #include "linux/fs.hpp"
 
 using std::ostream;
@@ -445,5 +447,195 @@ Try<uint64_t> weight(const string& cgroup)
 }
 
 } // namespace cpu {
+
+namespace devices {
+
+// Utility class to construct an eBPF program to whitelist or blacklist
+// select device accesses.
+class DeviceProgram
+{
+public:
+  DeviceProgram() : program{ebpf::Program(BPF_PROG_TYPE_CGROUP_DEVICE)}
+  {
+    program.append({
+      // The BPF_PROG_TYPE_CGROUP_DEVICE program takes in
+      // `struct bpf_cgroup_dev_ctx*` as input. We extract the fields into
+      // registers r2-5.
+      //
+      // The device type is encoded in the first 16 bits of `access_type` and
+      // the access type is encoded in the last 16 bits of `access_type`.
+      //
+      // r2: Type ('c', 'b', '?')
+      BPF_LDX_MEM(
+        BPF_W, BPF_REG_2, BPF_REG_1, offsetof(bpf_cgroup_dev_ctx, access_type)),
+      BPF_ALU32_IMM(BPF_AND, BPF_REG_2, 0xFFFF),
+      // r3: Access ('r', 'w', 'm')
+      BPF_LDX_MEM(BPF_W, BPF_REG_3, BPF_REG_1,
+        offsetof(bpf_cgroup_dev_ctx, access_type)),
+      BPF_ALU32_IMM(BPF_RSH, BPF_REG_3, 16),
+      // r4: Major Version
+      BPF_LDX_MEM(BPF_W, BPF_REG_4, BPF_REG_1,
+        offsetof(bpf_cgroup_dev_ctx, major)),
+      // r5: Minor Version
+      BPF_LDX_MEM(BPF_W, BPF_REG_5, BPF_REG_1,
+        offsetof(bpf_cgroup_dev_ctx, minor)),
+    });
+  }
+
+  Try<Nothing> allow(const Entry entry) { return addDevice(entry, true); }
+
+  Try<Nothing> deny(const Entry entry) { return addDevice(entry, false); }
+
+  ebpf::Program build()
+  {
+    if (!hasCatchAll) {
+      // Exit instructions.
+      // If no entry granted access, then deny the access.
+      program.append({
+        BPF_MOV64_IMM (BPF_REG_0, DENY_ACCESS),
+        BPF_EXIT_INSN(),
+      });
+    }
+    return program;
+  }
+
+private:
+  Try<Nothing> addDevice(const Entry entry, bool allow)
+  {
+    if (hasCatchAll) {
+      return Nothing();
+    }
+
+    // We create a block of bytecode with the format:
+    // 1. Major Version Check
+    // 2. Minor Version Check
+    // 3. Type Check
+    // 4. Access Check
+    // 5. Allow/Deny Access
+    //
+    // 6. NEXT BLOCK
+    //
+    // Either:
+    // 1. The device access is matched by (1,2,3,4) and the Allow/Deny access
+    //    block (5) is executed.
+    // 2. One of (1,2,3,4) does not match the requested access and we skip
+    //    to the next block (6).
+
+    const Entry::Selector& selector = entry.selector;
+    const Entry::Access& access = entry.access;
+
+    bool check_major = selector.major.isSome();
+    bool check_minor = selector.minor.isSome();
+    bool check_type = selector.type != Entry::Selector::Type::ALL;
+    bool check_access = !access.mknod || !access.read || !access.write;
+
+    // Number of instructions to the [NEXT BLOCK]. This is used if a check
+    // fails (meaning this entry does not apply) and we want to skip the
+    // subsequent checks.
+    short jmp_size = 1 + (check_major ? 1 : 0) + (check_minor ? 1 : 0) +
+                     (check_access ? 3 : 0) + (check_type ? 1 : 0);
+
+    // Check major version (r4) against entry.
+    if (check_major) {
+      program.append({
+        BPF_JMP_IMM(BPF_JNE, BPF_REG_4, (int)selector.major.get(), jmp_size),
+      });
+      --jmp_size;
+    }
+
+    // Check minor version (r5) against entry.
+    if (check_minor) {
+      program.append({
+        BPF_JMP_IMM(BPF_JNE, BPF_REG_5, (int)selector.minor.get(), jmp_size),
+      });
+      --jmp_size;
+    }
+
+    // Check type (r2) against entry.
+    if (check_type) {
+      int bpf_type = [selector]() {
+        switch (selector.type) {
+          case Entry::Selector::Type::BLOCK:     return BPF_DEVCG_DEV_BLOCK;
+          case Entry::Selector::Type::CHARACTER: return BPF_DEVCG_DEV_CHAR;
+          case Entry::Selector::Type::ALL:       UNREACHABLE();
+        }
+      }();
+
+      program.append({
+        BPF_JMP_IMM(BPF_JNE, BPF_REG_2, bpf_type, jmp_size),
+      });
+      --jmp_size;
+    }
+
+    // Check access (r3) against entry.
+    if (check_access) {
+      int bpf_access = 0;
+      bpf_access |= access.read ? BPF_DEVCG_ACC_READ : 0;
+      bpf_access |= access.write ? BPF_DEVCG_ACC_WRITE : 0;
+      bpf_access |= access.mknod ? BPF_DEVCG_ACC_MKNOD : 0;
+
+      program.append({
+        BPF_MOV32_REG(BPF_REG_1, BPF_REG_3),
+        BPF_ALU32_IMM(BPF_AND, BPF_REG_1, bpf_access),
+        BPF_JMP_REG(
+          BPF_JNE, BPF_REG_1, BPF_REG_3, static_cast<short>(jmp_size - 2)),
+      });
+      jmp_size -= 3;
+    }
+
+    if (!check_major && !check_minor && !check_type && !check_access) {
+      // The exit instructions as well as any additional device entries would
+      // generate unreachable blocks.
+      hasCatchAll = true;
+    }
+
+    // Allow/Deny access block.
+    program.append({
+      BPF_MOV64_IMM(BPF_REG_0, allow ? ALLOW_ACCESS : DENY_ACCESS),
+      BPF_EXIT_INSN(),
+    });
+
+    return Nothing();
+  }
+
+  ebpf::Program program;
+
+  // True if the program has a device entry that allows or denies ALL accesses.
+  // Such cases need to be specially handled because any instructions added
+  // after it will be unreachable, and thus will cause the eBPF verifier to
+  // reject the program.
+  bool hasCatchAll = false;
+
+  static const int ALLOW_ACCESS = 1;
+
+  static const int DENY_ACCESS = 0;
+};
+
+
+Try<Nothing> configure(
+    const string& cgroup,
+    const vector<Entry>& allow,
+    const vector<Entry>& deny)
+{
+  DeviceProgram program = DeviceProgram();
+  foreach (const Entry entry, allow) {
+    program.allow(entry);
+  }
+  foreach (const Entry entry, deny) {
+    program.deny(entry);
+  }
+
+  Try<Nothing> attach =
+    ebpf::cgroups2::attach(cgroups2::path(cgroup), program.build());
+  if (attach.isError()) {
+    return Error(
+        "Failed to attach BPF_PROG_TYPE_CGROUP_DEVICE program: " +
+        attach.error());
+  }
+
+  return Nothing();
+}
+
+} // namespace devices {
 
 } // namespace cgroups2 {
--- a/src/linux/cgroups2.hpp
+++ b/src/linux/cgroups2.hpp
@@ -24,6 +24,8 @@
 #include <stout/nothing.hpp>
 #include <stout/try.hpp>
 
+#include "linux/cgroups.hpp"
+
 namespace cgroups2 {
 
 // Root cgroup in the cgroup v2 hierarchy. Since the root cgroup has the same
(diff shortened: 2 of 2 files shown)


src/tests/containerizer/cgroups2_tests.cpp
```diff
@@ -59,6 +63,30 @@ TEST_F(Cgroups2Test, ROOT_CGROUPS2_AvailableSubsystems)
...
+  Try<vector<uint32_t>> attached = ebpf::cgroups2::attached(
+      cgroups2::ROOT_CGROUP);
+
+  ASSERT_SOME(attached);
+  EXPECT_EQ(1, (int) attached->size());
+
+  // Verify that we can't open /dev/null.
+  EXPECT_ERROR(os::open(os::DEV_NULL, O_RDWR));
+
+  EXPECT_SOME(ebpf::cgroups2::detach(cgroups2::ROOT_CGROUP, attached.get()[0]));
```
bmahler, line 84: Another reason we don't want to modify the root cgroup, if our test fails or is killed, we'll leave our device program attached?

bmahler, line 84: Prefer `*` or `->at(0)` over .get().

src/linux/cgroups2.hpp
```diff
@@ -69,6 +71,20 @@ Try<std::set<std::string>> enabled(const std::string& cgroup);
...
+
+using cgroups::devices::Entry;
+
+// Configure the device access permissions for the cgroup. These permissions
+// are hierarchical. I.e. if a parent cgroup does not allow an access then
+// 'this' cgroup will be denied access.
+Try<Nothing> configure(
+    const std::string& cgroup,
+    const std::vector<Entry> allow,
+    const std::vector<Entry> deny);
```
bmahler, line 84: We typically do not take a const non reference like this. The options are:

* const vector& (if we don't need to mutate it / store it)
* vector&& (if we want to force a move to avoid copying)
* vector (if we need to mutate it / store a mutated version of it, and want to support both move and const ref / copy)

src/linux/cgroups2.cpp
```diff
@@ -295,4 +296,152 @@ Try<set<string>> enabled(const string& cgroup)
 
 } // namespace controllers {
 
+namespace devices {
+
+const int ALLOW_ACCESS = 1;
+
+const int DENY_ACCESS = 0;
```
bmahler, line 303: Move these inside DeviceProgram?

src/linux/cgroups2.cpp
```diff
@@ -295,4 +296,152 @@ Try<set<string>> enabled(const string& cgroup)
...
+        // r2: Type ('c', 'b', '?')
+        BPF_LDX_MEM(BPF_W, BPF_REG_2, BPF_REG_1, 0),
+        BPF_ALU32_IMM(BPF_AND, BPF_REG_2, 0xFFFF),
+        // r3: Access ('r', 'w', 'm')
+        BPF_LDX_MEM(BPF_W, BPF_REG_3, BPF_REG_1, 0),
+        BPF_ALU32_IMM(BPF_RSH, BPF_REG_3, 16),
+        // r4: Major Version
+        BPF_LDX_MEM(BPF_W, BPF_REG_4, BPF_REG_1, 4),
+        // r5: Minor Version
+        BPF_LDX_MEM(BPF_W, BPF_REG_5, BPF_REG_1, 8)});
```
bmahler, line 322: quite hard to understand what this is doing, can you add some explanation that covers the byte format and the operations we're doing to extract the values out into separate registers?

bmahler, line 322: when using an initializer list {, you can indent by two, and close the } on its own line (if a trailing comma on the last item works as well, add one):

```
  BPF_LDX_MEM(BPF_W, BPF_REG_5, BPF_REG_1, 8),
});
```

src/linux/cgroups2.cpp
```diff
@@ -295,4 +296,152 @@ Try<set<string>> enabled(const string& cgroup)
...
+  }
+
+  Try<Nothing> allow(const Entry entry) { return addDevice(entry, true); }
+
+  Try<Nothing> deny(const Entry entry) { return addDevice(entry, false); }
+
+  Try<Nothing> addDevice(const Entry entry, bool allow)
```
bmahler, line 329: do you want this private or protected?

src/linux/cgroups2.cpp
```diff
@@ -295,4 +296,152 @@ Try<set<string>> enabled(const string& cgroup)
...
+  {
+    // We create a block of bytecode with the format:
+    // 1. [ Type Check          ] ⤵
+    // 2. [ Access Check        ] ⤵
+    // 3. [ Minor Version Check ] ⤵
+    // 4. [ Major Version Check ] ⤵
+    // 5. [ Allow/Deny Access   ]  ↓
+    //                             ↓
+    // 6. [ NEXT BLOCK          ] ←←
```
bmahler, line 338: can you remove the unicode characters? maybe this works, but I don't think we've added any non-ASCII to our current files.. have we?

bmahler, line 338: What is the NEXT BLOCK?

src/linux/cgroups2.cpp
```diff
@@ -295,4 +296,152 @@ Try<set<string>> enabled(const string& cgroup)
...
+
+  Try<Nothing> allow(const Entry entry) { return addDevice(entry, true); }
+
+  Try<Nothing> deny(const Entry entry) { return addDevice(entry, false); }
+
+  Try<Nothing> addDevice(const Entry entry, bool allow)
+  {
+    // We create a block of bytecode with the format:
+    // 1. [ Type Check          ] ⤵
+    // 2. [ Access Check        ] ⤵
```
bmahler, line 333: Hm.. why would we check access before we know that we're looking at the right device? i.e. that the device's type and major / minor match?

src/linux/cgroups2.cpp
```diff
@@ -295,4 +296,152 @@ Try<set<string>> enabled(const string& cgroup)
...
+    // 1. The device access is matched by (1,2,3,4) and the Allow/Deny access
+    //    block (5) is executed.
+    // 2. One of (1,2,3,4) does not match the requested access and we skip
+    //    to the next block (6).
+
+    // We only check attributes that are not 100% permissive:
+    // - Type   != all
+    // - Access != rnw
+    // - Minor  != None (AKA "any")
+    // - Major  != None (AKA "any")
```
bmahler, line 350: What does this mean? This comment isn't clear to an outsider IMO.

edit: after reading the rest of the code I now get this, it seems to me like this comment should be right on top of the check_ bools rather than one block above it:

```
    // The generated program only needs to perform checks when the selector / access
    // are specified as follows:
    bool check_type = selector.type != Entry::Selector::Type::ALL;
    bool check_access = !access.mknod || !access.read || !access.write;
    bool check_major = selector.major.isSome();
    bool check_minor = selector.minor.isSome();
```

src/linux/cgroups2.cpp
```diff
@@ -295,4 +296,152 @@ Try<set<string>> enabled(const string& cgroup)
...
+    bool check_type = selector.type != Entry::Selector::Type::ALL;
+    bool check_access = !access.mknod || !access.read || !access.write;
+    bool check_major = selector.major.isSome();
+    bool check_minor = selector.minor.isSome();
+
+    // Number of instructions to the [NEXT BLOCK]. This is used if a check
+    // fails (meaning this whitelist entry does not apply) and we want to
+    // skip the subsequent checks.
+    short jmp_size = 1 + (check_access ? 3 : 0) + (check_type ? 1 : 0) +
+                     (check_major ? 1 : 0) + (check_minor ? 1 : 0);
```
bmahler, line 363: where are these 3's and 1's coming from? can they be computed rather than burned in?

edit: I see now based on what gets appended below

src/linux/cgroups2.cpp
```diff
@@ -295,4 +296,152 @@ Try<set<string>> enabled(const string& cgroup)
...
+    // 2. One of (1,2,3,4) does not match the requested access and we skip
+    //    to the next block (6).
+
+    // We only check attributes that are not 100% permissive:
+    // - Type   != all
+    // - Access != rnw
+    // - Minor  != None (AKA "any")
+    // - Major  != None (AKA "any")
+    Entry::Selector selector = entry.selector;
+    Entry::Access access = entry.access;
```
bmahler, line 352: Feel free to take const refs here to indicate we're not modifying these?

src/linux/cgroups2.cpp
```diff
@@ -295,4 +296,152 @@ Try<set<string>> enabled(const string& cgroup)
...
+    // fails (meaning this whitelist entry does not apply) and we want to
+    // skip the subsequent checks.
+    short jmp_size = 1 + (check_access ? 3 : 0) + (check_type ? 1 : 0) +
+                     (check_major ? 1 : 0) + (check_minor ? 1 : 0);
+
+    // Check type (r2) against entry.
+    if (check_type) {
+      int bpf_type = selector.type == Entry::Selector::Type::BLOCK
+                       ? BPF_DEVCG_DEV_BLOCK
+                       : BPF_DEVCG_DEV_CHAR;
```
bmahler, line 369: can you use a switch on selector.type here instead with no default case? The compiler will then fail if someone adds a new selector.type enum value but this code isn't updated

src/linux/cgroups2.cpp
```diff
@@ -295,4 +296,152 @@ Try<set<string>> enabled(const string& cgroup)
...
+
+      program.append({ BPF_JMP_IMM(BPF_JNE, BPF_REG_2, bpf_type, jmp_size) });
+      --jmp_size;
+    }
+    // Check access (r3) against entry.
```
bmahler, line 374: newline between each new block of code here

src/linux/cgroups2.cpp
```diff
@@ -295,4 +296,152 @@ Try<set<string>> enabled(const string& cgroup)
...
+     program.append({
+          BPF_MOV32_REG (BPF_REG_1, BPF_REG_3),
+          BPF_ALU32_IMM (BPF_AND, BPF_REG_1, bpf_access),
+          BPF_JMP_REG (BPF_JNE, BPF_REG_1, BPF_REG_3, (short) (jmp_size - 2))});
+      jmp_size -= 3;
+    }
+    // Check major version (r4) against entry.
+    if (check_major) {
+      program.append({
+          BPF_JMP_IMM(BPF_JNE, BPF_REG_4, (int) selector.major.get(), jmp_size)});
```
bmahler, line 390: would be good to explain these instructions in general, e.g.

```
// Jump if major (r4) != selector.major.
```

Are we jumping *past* the first instruction of the allow / deny access block below? If so, it seems like it makes sense to separate out the append of the BPF_EXIT_INSN instruction and make it more clear in the comments here that we're jumping directly to that?

src/linux/cgroups2.cpp
```diff
@@ -295,4 +296,152 @@ Try<set<string>> enabled(const string& cgroup)
...
+
+  ebpf::Program build()
+  {
+    // Exit instructions.
+    // If no entry granted access, then deny the access.
+    program.append({
+        BPF_MOV64_IMM (BPF_REG_0, DENY_ACCESS),
+        BPF_EXIT_INSN()});
+
+    return program;
```
bmahler, line 416: are we expecting our programs to reach these instructions..? if so, when?

bmahler (commented):
nice! left some feedback below

src/tests/containerizer/cgroups2_tests.cpp
```diff
@@ -59,6 +63,30 @@ TEST_F(Cgroups2Test, ROOT_CGROUPS2_AvailableSubsystems)
...
+TEST_F(Cgroups2Test, ROOT_CGROUPS2_DeviceController)
+{
+  namespace devices = cgroups2::devices;
+
+  // Deny all access.
+  devices::Entry entry = devices::Entry::parse("a *:* rwm").get();
+
+  EXPECT_SOME(devices::configure(cgroups2::ROOT_CGROUP, {}, {entry}));
+  Try<vector<uint32_t>> attached = ebpf::cgroups2::attached(
+      cgroups2::ROOT_CGROUP);
```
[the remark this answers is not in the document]
DevinLeamy: Agreed. Will do :+1:

src/tests/containerizer/cgroups2_tests.cpp
```diff
@@ -59,6 +63,30 @@ TEST_F(Cgroups2Test, ROOT_CGROUPS2_AvailableSubsystems)
...
+
+  ASSERT_SOME(attached);
+  EXPECT_EQ(1, (int) attached->size());
```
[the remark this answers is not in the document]
DevinLeamy: The error message it produces is not very telling, but yeah the cast is required. I'll use `1u` instead :+1:

src/tests/containerizer/cgroups2_tests.cpp
DevinLeamy: Will update to create a new cgroup for testing.

DevinLeamy: Wasn't aware of `->at(0)`, will do that (`*attached[0]` doesn't work, unfortunately)

src/linux/cgroups2.hpp
DevinLeamy: Will update to be `const vector&` :+1:

src/linux/cgroups2.cpp
DevinLeamy: `private` :+1:

src/linux/cgroups2.cpp
DevinLeamy: Sure, I'll remove them.

DevinLeamy: The instruction _after_ the last instruction that is added by a call to `addDevice()`

src/linux/cgroups2.cpp
DevinLeamy: Since all of the checks need to pass, I considered the order to be somewhat irrelevant. Ordering is semantically the same, we'd just short-circuit.

src/linux/cgroups2.cpp
DevinLeamy: Will move the comment :+1:

src/linux/cgroups2.cpp
DevinLeamy: If no devices are added to a program. If we _don't_ have these instructions and you try and load the program without any device entries into the kernel, it will cause a validation error. It's protective.

src/tests/containerizer/cgroups2_tests.cpp
bmahler: it would need to be `(*attached)[0]` due to operator precedence rules IIRC

src/linux/cgroups2.cpp
bmahler: I understand that regardless of ordering, these all need to be checked and we jump out if any don't match, but it seems more natural to reason about checking in the same order that we'd be writing the logic if we were to code the logic if we weren't generating assembly, in which case I would imagine that the code would be written to first examine if this is the right device, then check access. It also seems a bit easier to reason about the access part if it resides closer to the final access code?

src/linux/cgroups2.cpp
bmahler: ah ok, in that case, can we generate this only when needed, for clarity?

src/linux/cgroups2.cpp
DevinLeamy: Actually, since it's only in the`*.cpp` file I'm keeping it public. I would have assumed that I could mark it public and still have access within `cgroups2.cpp`, but that doesn't look to be the case.

src/linux/cgroups2.cpp
DevinLeamy: Actually, reflecting, we reach this point just as described in the comment:
```
// If no entry granted access, then deny the access.
```

src/linux/cgroups2.cpp
DevinLeamy: Reordered.

src/linux/cgroups2.cpp
DevinLeamy: Make this more clear.

src/linux/cgroups2.hpp
```diff
@@ -75,6 +77,10 @@ Try<Nothing> move_process(const std::string& cgroup, pid_t pid);
 // /sys/fs/cgroup. E.g. For /sys/fs/cgroup/test, this will return "test".
 Try<std::string> cgroup(pid_t pid);
 
+
+// Get the absolute of a cgroup. Assumes to cgroup exists.
+std::string path(const std::string& cgroup);
```
bmahler, line 82: maybe add this in a different PR and replace all the existing path joins with this?

src/linux/cgroups2.cpp
```diff
@@ -442,4 +449,204 @@ Try<uint64_t> weight(const string& cgroup)
...
+  static const int ALLOW_ACCESS = 1;
+
+  static const int DENY_ACCESS = 0;
+};
+
+
+Try<Nothing> configure(
+  const string& cgroup,
+  const vector<Entry>& allow,
+  const vector<Entry>& deny)
```
bmahler, line 630: 4 spaces here?

src/linux/cgroups2.cpp
```diff
@@ -442,4 +449,204 @@ Try<uint64_t> weight(const string& cgroup)
...
+  }
+  foreach (const Entry entry, deny) {
+    program.deny(entry);
+  }
+
+  Try<Nothing> attach =
+    ebpf::cgroups2::attach(cgroups2::path(cgroup), program.build());
+  if (attach.isError()) {
+    return Error(
+        "Failed to attach Device Controller program: " + attach.error());
```
bmahler, line 644: maybe "cgroup device program" or BPF_PROG_TYPE_CGROUP_DEVICE?

src/tests/containerizer/cgroups2_tests.cpp
```diff
@@ -115,6 +119,48 @@ TEST_F(Cgroups2Test, ROOT_CGROUPS2_AssignProcessToCgroup)
...
+TEST_F(Cgroups2Test, ROOT_CGROUPS2_DeviceController)
+{
+  namespace devices = cgroups2::devices;
+
+  ASSERT_SOME(cgroups2::create(TEST_CGROUP));
+  string absolutePath = cgroups2::path(TEST_CGROUP);
+
+  // Deny all access.
+  Try<devices::Entry> entry = devices::Entry::parse("c *:* rwm");
+  ASSERT_SOME(entry);
```
bmahler, line 132: would be good to get more coverage of the logic here, rather than only testing a single deny all character devices?

PR #520, opened 2024-03-18
[cgroups2] Introduces cgroups2::create() to create new cgroups.

DevinLeamy:
Adds `cgroups2::create` to the cgroups2 API, allowing new cgroups to be created.

### DIFF
--- a/src/linux/cgroups2.cpp
+++ b/src/linux/cgroups2.cpp
@@ -242,6 +242,23 @@ Try<Nothing> unmount()
   return Nothing();
 }
 
+
+Try<Nothing> create(const string& cgroup, bool recursive)
+{
+  string absolutePath = path::join(MOUNT_POINT, cgroup);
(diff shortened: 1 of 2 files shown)


src/linux/cgroups2.cpp
```diff
@@ -242,6 +242,23 @@ Try<Nothing> unmount()
   return Nothing();
 }
 
+
+Try<Nothing> create(const string& cgroup, bool recursive)
+{
+  string absolutePath = path::join(MOUNT_POINT, cgroup);
+  if (os::exists(absolutePath)) {
+    return Error("A cgroup at '" + absolutePath + "' already exists");
+  }
```
bmahler, line 251: Do we need this? mkdir should give us EEXIST in this case?

bmahler (approved):
I'll get this landed with the below change.

PR #522, opened 2024-03-19
[cgroups2] Introduces the cpu namespace and cpu::weight() API

DevinLeamy:
Introduces the `cpu` namespace for cgroups2, and expose an API to set
the CPU weight for a cgroup.

In cgroups v2, CPU cycles are allocated based on weights that are assigned
to cgroups, using the "cpu.weight" control. Specifically, a cgroup has access
to 100(weight/total_weight)% of CPU access.

### DIFF
--- a/src/linux/cgroups2.cpp
+++ b/src/linux/cgroups2.cpp
@@ -382,4 +382,22 @@ Try<set<string>> enabled(const string& cgroup)
 
 } // namespace controllers {
 
+namespace cpu {
+
+namespace control {
+
+const std::string IDLE = "cpu.idle";
+const std::string MAX = "cpu.max";
+const std::string MAX_BURST = "cpu.max.burst";
+const std::string PRESSURE = "cpu.pressure";
+const std::string STATS = "cpu.stat";
+const std::string UCLAMP_MAX = "cpu.uclamp.max";
+const std::string UCLAMP_MIN = "cpu.uclamp.min";
+const std::string WEIGHT = "cpu.weight";
+const std::string WEIGHT_NICE = "cpu.weight.nice";
+
+} // namespace control {
+
+} // namespace cpu {
+
 } // namespace cgroups2 {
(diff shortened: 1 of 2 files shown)


src/linux/cgroups2.hpp
```diff
@@ -95,6 +95,19 @@ Try<std::set<std::string>> enabled(const std::string& cgroup);
...
 } // namespace controllers {
 
+namespace cpu {
+
+// Set the weight for a cgroup.
+// Cannot be used for the root cgroup.
+Try<Nothing> weight(const std::string& cgroup, uint32_t weight);
+
+
+// Determine the weight a cgroup.
```
bmahler, line 105: Ditto here.

src/linux/cgroups2.cpp
```diff
@@ -382,4 +404,42 @@ Try<set<string>> enabled(const string& cgroup)
...
+
+} // namespace control {
+
+Try<Nothing> weight(const string& cgroup, uint32_t weight)
+{
+  if (cgroup == ROOT_CGROUP) {
+    return Error("Operation does not exist for the root cgroup");
+  }
+
+  return cgroups2::write(cgroup, cpu::control::WEIGHT, (uint64_t) weight);
```
bmahler, line 429: this is a C style cast, we would use static_cast here to check at compile time that it works, however we probably don't need to be casting since we don't have an integer type that holds the needed range anyway

bmahler, line 429: this is a C style cast, we would use static_cast here to check at compile time that it works, however we probably don't need to be casting since we don't have an integer type that holds the needed range anyway

src/linux/cgroups2.cpp
```diff
@@ -382,4 +404,42 @@ Try<set<string>> enabled(const string& cgroup)
...
+const std::string UCLAMP_MIN = "cpu.uclamp.min";
+const std::string WEIGHT = "cpu.weight";
+const std::string WEIGHT_NICE = "cpu.weight.nice";
+
+} // namespace control {
+
+Try<Nothing> weight(const string& cgroup, uint32_t weight)
+{
+  if (cgroup == ROOT_CGROUP) {
+    return Error("Operation does not exist for the root cgroup");
```
bmahler, line 426: maybe "not supported" is more idiomatic here

PR #523, opened 2024-03-19
[cgroups2] Introduces API to move a process into a cgroup, by PID.

DevinLeamy:
A process can be moved into a cgroup by writing its process id into
the cgroup's `cgroup.procs` file. When a process is moved, all its
threads move with it.

In this PR we introduce `cgroups2::move_process` to add a process to
a cgroup, `cgroups2::cgroup` to check which cgroup a process is a
member of, and test the introduced functionality.

### DIFF
--- a/src/linux/cgroups2.cpp
+++ b/src/linux/cgroups2.cpp
@@ -274,6 +274,46 @@ Try<Nothing> destroy(const string& cgroup)
   return Nothing();
 }
 
+
+Try<Nothing> move_process(const string& cgroup, pid_t pid)
+{
+  const string absolutePath = path::join(MOUNT_POINT, cgroup);
+
+  if (!os::exists(absolutePath)) {
+    return Error("There does not exist a cgroup at '" + absolutePath + "'");
+  }
+
+  return cgroups2::write(cgroup, control::PROCESSES, std::to_string(pid));
+}
+
+
+Try<string> cgroup(pid_t pid)
+{
+  // A process's cgroup membership is listed in '/proc/{pid}/cgroup'.
+  //
+  // The format, e.g for the cgroup `/sys/fs/cgroup/foo/bar`, is:
+  // """"
+  // 0::/foo/bar
+  // or
+  // 0::/foo/bar (deleted)
+  // """
+  const string& cgroupFile = path::join("/proc", std::to_string(pid), "cgroup");
+  if (!os::exists(cgroupFile)) {
+    return Error("File '" + cgroupFile + "' does not exist");
+  }
+
+  Try<string> content = os::read(cgroupFile);
+  if (content.isError()) {
+    return Error("Failed to read '" + cgroupFile + "'");
+  }
+
+  string parsed = strings::remove(
+      strings::trim(*content), "0::/", strings::Mode::PREFIX);
+  parsed = strings::remove(parsed, " (deleted)", strings::Mode::SUFFIX);
+
+  return parsed;
+}
+
 namespace controllers {
 
 Try<set<string>> available(const string& cgroup)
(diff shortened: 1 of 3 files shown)


src/linux/cgroups2.cpp
```diff
@@ -265,7 +265,7 @@ Try<Nothing> destroy(const string& cgroup)
     return Error("There does not exist a cgroup at '" + absolutePath + "'");
   }
 
-  Try<Nothing> rmdir = os::rmdir(absolutePath);
+  Try<Nothing> rmdir = os::rmdir(absolutePath, false);
```
DevinLeamy, line 268: This PR has two commits. This change (and only this change) is from the first commit, where I explain why this was introduced. TLDR: make `os::rmdir` use `rmdir` instead of `rm`.

src/linux/cgroups2.cpp
```diff
@@ -274,6 +274,46 @@ Try<Nothing> destroy(const string& cgroup)
...
+
+Try<Nothing> move_process(const string& cgroup, pid_t pid)
+{
+  const string absolutePath = path::join(MOUNT_POINT, cgroup);
+
+  if (!os::exists(absolutePath)) {
+    return Error("There does not exist a cgroup at '" + absolutePath + "'");
+  }
+
+  return cgroups2::write(cgroup, control::PROCESSES, std::to_string(pid));
```
bmahler, line 286: we use stringify() for stringification, ditto below

src/linux/cgroups2.cpp
```diff
@@ -274,6 +274,46 @@ Try<Nothing> destroy(const string& cgroup)
...
+  // 0::/foo/bar (deleted)
+  // """
+  const string& cgroupFile = path::join("/proc", std::to_string(pid), "cgroup");
+  if (!os::exists(cgroupFile)) {
+    return Error("File '" + cgroupFile + "' does not exist");
+  }
+
+  Try<string> content = os::read(cgroupFile);
+  if (content.isError()) {
+    return Error("Failed to read '" + cgroupFile + "'");
```
bmahler, line 307: include the read error in the message here?

src/linux/cgroups2.cpp
```diff
@@ -274,6 +274,46 @@ Try<Nothing> destroy(const string& cgroup)
...
+  }
+
+  string parsed = strings::remove(
+      strings::trim(*content), "0::/", strings::Mode::PREFIX);
+  parsed = strings::remove(parsed, " (deleted)", strings::Mode::SUFFIX);
```
bmahler, line 312: hm.. seems like we should return an error if the pid belongs to a v1 cgroup and therefore we find the wrong format in the file? Rather than return a string like:

"5:cpuacct,cpu,cpuset:/daemons"

which is the current behavior in this case

src/tests/containerizer/cgroups2_tests.cpp
```diff
@@ -59,6 +63,35 @@ TEST_F(Cgroups2Test, ROOT_CGROUPS2_AvailableSubsystems)
   }
 }
 
+
+TEST_F(Cgroups2Test, ROOT_CGROUPS2_AssignProcessToCgroup)
+{
+  string cgroup = "test";
+  ASSERT_SOME(cgroups2::create(cgroup));
```
bmahler, line 70: We need to add cleanup logic into the Cgroups2Test fixture, otherwise if this test leaves early nothing cleans up the test cgroup, can you follow up on this in a separate PR?

bmahler (approved):
I can land this with local edits, but I did not take care of the test fixture doing cleanup, can you follow up on that?

PR #525, opened 2024-03-19
[cgroups2] Introduces cgroups2::exists to check if a cgroup exists.

DevinLeamy:
If you call `cgroups2::destroy` on a cgroup that does not exist, it will throw an error. This is expected. However, it means that the caller needs to know whether a cgroup exists. This PR introduces `cgroups2::exists`, allowing the caller find out if a cgroup exists.

This is useful, for example, for test fixtures where cgroups from a previous run of a test may or may not be cleaned up, depending on whether or not the test was successful.

### DIFF
--- a/src/linux/cgroups2.cpp
+++ b/src/linux/cgroups2.cpp
@@ -243,6 +243,12 @@ Try<Nothing> unmount()
 }
 
 
+bool exists(const string& cgroup)
+{
+  return os::exists(path::join(MOUNT_POINT, cgroup));
+}
+
+
 Try<Nothing> create(const string& cgroup, bool recursive)
 {
   const string absolutePath = path::join(MOUNT_POINT, cgroup);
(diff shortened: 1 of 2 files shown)


src/linux/cgroups2.cpp
```diff
@@ -259,12 +265,11 @@ Try<Nothing> create(const string& cgroup, bool recursive)
...
 {
-  const string absolutePath = path::join(MOUNT_POINT, cgroup);
-
-  if (!os::exists(absolutePath)) {
-    return Error("There does not exist a cgroup at '" + absolutePath + "'");
+  if (!cgroups2::exists(cgroup)) {
+    return Error("Cgroup '" + cgroup + "' does not exist");
   }
 
+  const string absolutePath = path::join(MOUNT_POINT, cgroup);
```
DevinLeamy, line 272: Can remove this change, but I think it's relevant because we're replacing the if statement check if an equivalent but more readable check.

PR #528, opened 2024-03-20
[cgroups2] Test fixture to cleanup `TEST_CGROUP`.

DevinLeamy:
Many cgroups tests require creating a child cgroup and then updating control files within that cgroup. To avoid requiring test writers to manually cleanup and conditionally create `TEST_CGROUP`, we add cleanup logic to the setup and teardown of the `Cgroups2Test` test fixture.

### DIFF
--- a/src/tests/containerizer/cgroups2_tests.cpp
+++ b/src/tests/containerizer/cgroups2_tests.cpp
@@ -33,7 +33,29 @@ namespace mesos {
 namespace internal {
 namespace tests {
 
-class Cgroups2Test : public TemporaryDirectoryTest {};
+const string TEST_CGROUP = "test";
+
+// Cgroups2 test fixture that ensures `TEST_CGROUP` does not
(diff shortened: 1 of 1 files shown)


src/tests/containerizer/cgroups2_tests.cpp
```diff
@@ -33,7 +33,29 @@ namespace mesos {
...
 
-class Cgroups2Test : public TemporaryDirectoryTest {};
+const string TEST_CGROUP = "test";
+
+// Cgroups2 test fixture that ensures `TEST_CGROUP` does not
+// exist before and after the test is run.
+class Cgroups2Test : public TemporaryDirectoryTest
+{
+public:
+  static void SetUpTestCase()
```
bmahler, line 43: this looks like it should be `void SetUp() override` and below `void TearDown() override`, which will run at the start and end of each test, as is, these statics only run at the start and end of the fixture

PR #529, opened 2024-03-21
[cgroups2] Add cgroups2::path to get the path of a cgroup.

DevinLeamy:
Introduces a utility to get the absolute path of a cgroup and refactors code to use it.

### DIFF
--- a/src/linux/cgroups2.cpp
+++ b/src/linux/cgroups2.cpp
@@ -147,7 +147,7 @@ std::ostream& operator<<(std::ostream& stream, const State& state)
 template <>
 Try<string> read(const string& cgroup, const string& control)
 {
-  return os::read(path::join(cgroups2::MOUNT_POINT, cgroup, control));
+  return os::read(path::join(cgroups2::path(cgroup), control));
 }
 
 
@@ -169,7 +169,7 @@ Try<Nothing> write(
     const string& control,
     const string& value)
 {
-  return os::write(path::join(cgroups2::MOUNT_POINT, cgroup, control), value);
+  return os::write(path::join(cgroups2::path(cgroup), control), value);
 }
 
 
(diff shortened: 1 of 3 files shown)


src/tests/containerizer/cgroups2_tests.cpp
```diff
@@ -115,6 +115,14 @@ TEST_F(Cgroups2Test, ROOT_CGROUPS2_AssignProcessToCgroup)
   AWAIT_EXPECT_WTERMSIG_EQ(SIGKILL, process::reap(pid));
 }
 
+
+TEST_F(Cgroups2Test, CGROUPS2_Path)
+{
+  EXPECT_EQ("/sys/fs/cgroup/", cgroups2::path(cgroups2::ROOT_CGROUP));
+  EXPECT_EQ("/sys/fs/cgroup/foo", cgroups2::path("foo"));
+  EXPECT_EQ("/sys/fs/cgroup/foo/bar", cgroups2::path("foo/bar"));
```
DevinLeamy, line 123: The matches the behavior of python3s `os.path.join`.

src/linux/cgroups2.hpp
```diff
@@ -75,6 +75,10 @@ Try<Nothing> move_process(const std::string& cgroup, pid_t pid);
 // /sys/fs/cgroup. E.g. For /sys/fs/cgroup/test, this will return "test".
 Try<std::string> cgroup(pid_t pid);
 
+
+// Get the absolute of a cgroup. Assumes to cgroup exists.
```
bmahler, line 79: s/to/the/

what does this assumption entail though?

src/tests/containerizer/cgroups2_tests.cpp
```diff
@@ -115,6 +115,14 @@ TEST_F(Cgroups2Test, ROOT_CGROUPS2_AssignProcessToCgroup)
   AWAIT_EXPECT_WTERMSIG_EQ(SIGKILL, process::reap(pid));
 }
 
+
+TEST_F(Cgroups2Test, CGROUPS2_Path)
+{
+  EXPECT_EQ("/sys/fs/cgroup/", cgroups2::path(cgroups2::ROOT_CGROUP));
```
bmahler, line 121: I guess this is indicating that you have two ways of calling this function?

1. Pass a "cgroup" relative to the hierarchy, you get back the absolute path.
2. Pass a "cgroup" using an absolute path, the absolute path must include the hierarchy already in this case, e.g. if you pass "/foo", you'll only get back "/foo", not "/sys/fs/cgroup/foo".

src/linux/cgroups2.cpp
```diff
@@ -309,10 +307,10 @@ Try<Nothing> destroy(const string& cgroup)
 
 Try<Nothing> move_process(const string& cgroup, pid_t pid)
 {
-  const string absolutePath = path::join(MOUNT_POINT, cgroup);
+  const string& path = cgroups2::path(cgroup);
```
bmahler, line 310: we can remove the ref here to be consistent with the other calls

PR #530, opened 2024-03-21
[cgroups2] Hotfix for controllers::enable/available.

DevinLeamy:
"cgroup.controllers" and "cgroup.subtree_control" control files are terminated by a newline, meaning that when we split the contents by " " we get a trailing empty value, which causes downstream errors when writing.

### DIFF
--- a/src/linux/cgroups2.cpp
+++ b/src/linux/cgroups2.cpp
@@ -112,9 +112,14 @@ struct State
     return _enabled.find(controller) != _enabled.end();
   }
 
-  static State parse(const string& contents)
+  static State parse(const string& _contents)
   {
     State control;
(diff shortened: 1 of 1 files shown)


bmahler (commented):
Ok, so the description is wrong now? Should be something more like:

cgroup.controllers" and "cgroup.subtree_control" control files can be empty,
and always end with a newline, following the "space separated values" format
for control files, e.g.:
```
\n
```
or
```
VAL0 VAL1\n
```

in the first case, `strings::split(s, " ")` returns `["\n"]` and in the second case,
we're getting a newline character on our last split item. We need to trim and
check for empty-ness to handle this correctly

PR #531, opened 2024-03-21
[cgroups2] Introduces interface to obtain CPU usage statistics.

DevinLeamy:
The "cpu.stat" control file contains statistics about CPU usage. We capture a snapshot of "cpu.stat" inside of a `cpu::Stats` object and expose it through `cgroups2::cpu::stats`.

### DIFF
--- a/src/linux/cgroups2.cpp
+++ b/src/linux/cgroups2.cpp
@@ -465,6 +465,53 @@ const std::string UCLAMP_MIN = "cpu.uclamp.min";
 const std::string WEIGHT = "cpu.weight";
 const std::string WEIGHT_NICE = "cpu.weight.nice";
 
+namespace stat {
+
+Try<Stats> parse(const string& content)
+{
+  const vector<string>& lines = strings::split(content, "\n");
+  cpu::Stats stats;
+
+  foreach (const string& line, lines) {
+    const string trimmed = strings::trim(line);
+    if (trimmed.empty()) {
+      continue;
+    }
+
+    vector<string> tokens = strings::split(trimmed, " ");
+    if (tokens.size() != 2) {
+      return Error(
+          "Invalid line format in 'cpu.stat' expected "
+          "<key> <value> received: '" + trimmed + "'");
+    }
+
+    string field = tokens[0];
+    string value = tokens[1];
+
+    Try<Duration> duration = Duration::parse(value + "us");
+    Try<uint64_t> number = numify<uint64_t>(value);
+
+    // Duration::parse will succeed if `value` is a number, so we only check
+    // that `number` parsed successfully.
+    if (number.isError()) {
+      return Error("Failed to parse '" + field + "': " + number.error());
+    }
+
+    if (field == "usage_usec") { stats.usage = *duration; }
+    else if (field == "user_usec") { stats.user_time = *duration; }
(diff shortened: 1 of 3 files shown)


src/tests/containerizer/cgroups2_tests.cpp
```diff
@@ -62,6 +62,26 @@ class Cgroups2Test : public TemporaryDirectoryTest
...
+protected:
+  void SetUp() override
+  {
+    Cgroups2Test::SetUp();
+
+    ASSERT_SOME(cgroups2::create(TEST_CGROUP));
+
+    // Enable "cpu" in the root cgroup, so it's available to the test cgroup.
+    // TODO(dleamy): Disable "cpu" on tear down, if it wasn't initially enabled.
+    ASSERT_SOME(cgroups2::controllers::enable(cgroups2::ROOT_CGROUP, {"cpu"}));
```
DevinLeamy, line 77: Happy to introduce a `cgroups2::controllers::disable` before landing this 👍

src/linux/cgroups2.cpp
```diff
@@ -465,6 +465,53 @@ const std::string UCLAMP_MIN = "cpu.uclamp.min";
...
+
+    vector<string> tokens = strings::split(trimmed, " ");
+    if (tokens.size() != 2) {
+      return Error(
+          "Invalid line format in 'cpu.stat' expected "
+          "<key> <value> received: '" + trimmed + "'");
+    }
+
+    string field = tokens[0];
+    string value = tokens[1];
```
bmahler, line 489: no need to copy these

src/linux/cgroups2.cpp
```diff
@@ -465,6 +465,53 @@ const std::string UCLAMP_MIN = "cpu.uclamp.min";
...
+
+    Try<Duration> duration = Duration::parse(value + "us");
```
bmahler, line 491: we can just construct Duration using `Duration d = Microseconds(static_cast<int64_t>(*number))`

src/linux/cgroups2.cpp
```diff
@@ -473,7 +520,7 @@ Try<Nothing> weight(const string& cgroup, uint64_t weight)
     return Error("Operation not supported for the root cgroup");
   }
 
-  return cgroups2::write(cgroup, cpu::control::WEIGHT, weight);
+  return cgroups2::write<uint64_t>(cgroup, cpu::control::WEIGHT, weight);
```
bmahler, line 523: whoops?

src/tests/containerizer/cgroups2_tests.cpp
```diff
@@ -134,6 +154,12 @@ TEST_F(Cgroups2Test, ROOT_CGROUPS2_AssignProcesses)
   AWAIT_EXPECT_WTERMSIG_EQ(SIGKILL, process::reap(pid));
 }
 
+
+TEST_F(Cgroups2CpuTest, ROOT_CGROUPS2_CpuStats)
+{
+  ASSERT_SOME(cgroups2::cpu::stats(TEST_CGROUP));
+}
+
```
bmahler, line 162: Instead of adding a new fixture, we can keep the same fixture but add a function that:
* enables requested controllers
* remembers the side effect
* disables the controllers that were added, during TearDown

PR #532, opened 2024-03-21
[cgroups2] Introduces an interface to read and assign processes.

DevinLeamy:
`cgroup.procs` lists all of the processes inside of a cgroup, one per line. There may be duplicates. A single process can be moved into a cgroup by writing its process ID into `cgroup.procs`.

Here we introduce:
- Reading the process ids of the processes inside of a cgroup, and
- Assigning a process to a cgroup by process ID.

### DIFF
--- a/src/linux/cgroups2.cpp
+++ b/src/linux/cgroups2.cpp
@@ -312,12 +312,10 @@ Try<Nothing> destroy(const string& cgroup)
 }
 
 
-Try<Nothing> move_process(const string& cgroup, pid_t pid)
+Try<Nothing> assign(const string& cgroup, pid_t pid)
 {
-  const string path = cgroups2::path(cgroup);
-
-  if (!os::exists(path)) {
-    return Error("There does not exist a cgroup at '" + path + "'");
+  if (!cgroups2::exists(cgroup)) {
+    return Error("Cgroup '" + cgroup + "' does not exist");
   }
 
   return cgroups2::write(cgroup, control::PROCESSES, stringify(pid));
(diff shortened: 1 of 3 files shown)


src/linux/cgroups2.hpp
```diff
@@ -76,6 +76,15 @@ Try<Nothing> move_process(const std::string& cgroup, pid_t pid);
 Try<std::string> cgroup(pid_t pid);
 
 
+// Get the processes inside of a cgroup.
+Try<std::set<pid_t>> processes(const std::string& cgroup);
+
+
+// Assign a process to a cgroup, by PID. This removes the process from its
+// current cgroup.
+Try<Nothing> assign(const std::string& cgroup, pid_t pid);
```
bmahler, line 85: there's the same function already with cgroups2::move_process..? assign or assign_process does seem like a better name, but did you mean to add two of the same function?

src/tests/containerizer/cgroups2_tests.cpp
```diff
@@ -123,6 +123,43 @@ TEST_F(Cgroups2Test, CGROUPS2_Path)
   EXPECT_EQ("/sys/fs/cgroup/foo/bar", cgroups2::path("foo/bar"));
 }
 
+
+TEST_F(Cgroups2Test, CGROUPS_Path)
```
bmahler, line 127: This test is about cgroups2::processes and process assignment, not path?

If this is fixed to be s/CGROUPS/CGROUPS2/ then this clashes with the existing CGROUPS2_Path test?

There's already ROOT_CGROUPS2_AssignProcessToCgroup which seems to be about the same as this test, minus cgroups2::processes?

bmahler (changes requested):
There's already cgroups2::move_process which does the same thing as cgroups2::assign..?

src/tests/containerizer/cgroups2_tests.cpp
DevinLeamy: Removed the `ROOT_CGROUPS2_AssignProcessToCgroup` test and updated the test name.

src/linux/cgroups2.cpp
```diff
@@ -350,6 +350,47 @@ Try<string> cgroup(pid_t pid)
...
+  }
+
+  string trimmed = strings::trim(*contents);
+  if (trimmed.empty()) {
+    return set<pid_t>();
+  }
+
+  set<pid_t> pids;
+  foreach (const string& _pid, strings::split(strings::trim(*contents), "\n")) {
+    Try<pid_t> pid = numify<pid_t>(strings::trim(_pid));
```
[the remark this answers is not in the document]
DevinLeamy: Was doing it to be "safe". Removed.

PR #535, opened 2024-03-26
[cgroups2] Introduce cgroups v2 isolator process.

DevinLeamy:
Introduces `Cgroups2IsolatorProcess`, a new isolator process which
uses cgroups v2. Cgroups v2 equivalent of `CgroupsIsolatorProcess`.

We introduce a new class, rather than adapting the existing
`CgroupsIsolatorProcess`, because the two isolators need to behave
markedly differently.

Specifically, the v1 isolator:
- Maintains a mapping from the cgroup hierarchy where a controller is
  mounted to the `Subsystem` that manages it.
- Mounts cgroups hierarchies for subsystems, as they are required.
- Performs (virtually) all cgroups operations taking into account which
  hierarchy a subsystem and container is a member of.

all of which are not relevant in the context of cgroup v2's unified
hierarchy.

### DIFF
--- a/src/slave/containerizer/mesos/isolators/cgroups2/cgroups2.hpp
+++ b/src/slave/containerizer/mesos/isolators/cgroups2/cgroups2.hpp
@@ -1,0 +1,59 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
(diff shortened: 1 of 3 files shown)


src/slave/containerizer/mesos/isolators/cgroups2/cgroups2.hpp
```diff
@@ -0,0 +1,59 @@
...
+
+  ~Cgroups2IsolatorProcess() override;
+
+  bool supportsNesting() override;
+
+  bool supportsStandalone() override;
+
+private:
+  Cgroups2IsolatorProcess(
+      const hashmap<std::string, process::Owned<Subsystem>>& _subsystems);
```
DevinLeamy, line 49: Will remove references to `Subsystem`s once a `Controller` abstraction is created.

PR #536, opened 2024-03-26
[cgroups2] Introduce cgroups2::disable to disable a controller.

DevinLeamy:
Allows a controller to be disabled on a cgroup. A test is introduced.

We additionally move the read and write utilities so that they are accessible when we read the cgroup.subtree_control control file.

### DIFF
--- a/src/tests/containerizer/cgroups2_tests.cpp
+++ b/src/tests/containerizer/cgroups2_tests.cpp
@@ -22,6 +22,7 @@
 #include <process/gtest.hpp>
 
 #include <stout/exit.hpp>
+#include <stout/foreach.hpp>
 #include <stout/set.hpp>
 #include <stout/tests/utils.hpp>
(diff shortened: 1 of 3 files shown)


src/tests/containerizer/cgroups2_tests.cpp
```diff
@@ -177,6 +179,35 @@ TEST_F(Cgroups2Test, ROOT_CGROUPS2_CpuStats)
   ASSERT_SOME(cgroups2::cpu::stats(TEST_CGROUP));
 }
 
+
+TEST_F(Cgroups2Test, ROOT_CGROUPS2_EnableAndDisable)
+{
+  ASSERT_SOME(cgroups2::create(TEST_CGROUP));
```
bmahler, line 185: we want to call enable_controllers here to ensure the root cgroup has the cpu controller enabled, right?

PR #537, opened 2024-03-27
[cgroups2] Added path helpers that define the structure of the cgroup2 filesystem

DevinLeamy:
Because of the unified hierarchy and internal processes constraint in cgroups v2, a new cgroup directory structure is used for cgroups in cgroups v2.

Here we introduce paths helpers and lay out the directory structure of the cgroup2 filesystem, used in cgroups v2.

### DIFF
--- a/src/slave/containerizer/mesos/paths.cpp
+++ b/src/slave/containerizer/mesos/paths.cpp
@@ -654,6 +654,30 @@ Option<ContainerID> parseCgroupPath(
   return current;
 }
 
+namespace cgroups2 {
+
+string agent(const string& root, bool leaf)
+{
+  if (leaf) { return path::join(root, "agent", "leaf"); }
+  else      { return path::join(root, "agent"); }
+}
+
+
+string container(
+    const string& root,
+    const ContainerID& containerId,
+    bool leaf)
+{
+  const string& container = containerizer::paths::buildPath(
+      containerId,
+      CGROUP_SEPARATOR,
+      containerizer::paths::JOIN);
+
+  if (leaf) { return path::join(root, container, "leaf"); }
+  else      { return path::join(root, container);}
+}
+
+} // namespace cgroups2 {
 } // namespace paths {
 } // namespace containerizer {
 } // namespace slave {
--- a/src/slave/containerizer/mesos/paths.hpp
+++ b/src/slave/containerizer/mesos/paths.hpp
@@ -286,6 +286,44 @@ Option<ContainerID> parseCgroupPath(
     const std::string& cgroupsRoot,
     const std::string& cgroup);
 
+// All cgroups v2 paths are either leaf or non-leaf paths. Leaf paths end
+// in `/leaf`, contain processes, and may impose resource constraints. Non-leaf
+// paths do not contain processes and may impose resource constraints.
+// This is done to avoid the restrictions of the "Internal Process Contraint".
+// More information: https://docs.kernel.org/admin-guide/cgroup-v2.html#no-internal-process-constraint
+//
+// The cgroup2 file system has the structure:
+// <MOUNT>/<root>/agent                    Mesos Agent constraints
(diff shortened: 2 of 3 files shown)


src/slave/containerizer/mesos/paths.cpp
```diff
@@ -654,6 +654,30 @@ Option<ContainerID> parseCgroupPath(
   return current;
 }
 
+namespace cgroups2 {
+
+string agent(const string& root, bool leaf)
+{
+  if (leaf) { return path::join(root, "agent", "leaf"); }
+  else      { return path::join(root, "agent"); }
```
DevinLeamy, line 662: IMO this looks nicer than using the ternary operator or removing the `else`, because the paths line up. Ditto below.

src/slave/containerizer/mesos/paths.cpp
```diff
@@ -654,6 +654,30 @@ Option<ContainerID> parseCgroupPath(
   return current;
 }
 
+namespace cgroups2 {
+
```
DevinLeamy, line 658: Could have additionally added:
```cpp
// Path to the top-level Mesos cgroup. All global resource constraints
// are impose from this cgroup.
std::string root(const std::string& root)
{
  return root;
}
```

But given that it takes the flag called `cgroups_root` and simply just returns its parameter I didn't see much of a point.

src/slave/containerizer/mesos/paths.hpp
```diff
@@ -286,6 +286,44 @@ Option<ContainerID> parseCgroupPath(
...
+// <MOUNT>/<root>/<id>                     Container constraints
+// <MOUNT>/<root>/<id>/leaf                Container process
+//
+// <MOUNT>/<root>/<id>/mesos/<id2>         Nested container constraints
+// <MOUNT>/<root>/<id>/mesos/<id2>/leaf    Nested container process
+//
+// For every new level of nesting, `/mesos/<idN>` is added to the path.
+// Note: We assume these nested containers are not configured to "share"
+//       cgroups with their parent, in which case they would use the same cgroup
+//       paths as their parent.
```
bmahler, line 308: can probably remove this? I think in v2 the share_cgroups flag should probably be aliased to a new name like shared_isolation, since:

* this is what the user cares about, and
* creating the cgroups lets us get stats across the executor, nested containers, even though they aren't isolated from each other, which is nice

src/tests/containerizer/mesos_containerizer_paths_tests.cpp
```diff
@@ -95,6 +95,51 @@ TEST(MesosContainerizerPathsTest, BuildPathJoinMode)
       path::join(parent.value(), separator, child.value()));
 }
 
+TEST(MesosContainerizerPathsTest, CGROUPS2_Cgroups2Paths)
+{
+    namespace cgroups2 = mesos::internal::slave::containerizer::paths::cgroups2;
```
bmahler, line 100: indentation?

src/tests/containerizer/mesos_containerizer_paths_tests.cpp
```diff
@@ -95,6 +95,51 @@ TEST(MesosContainerizerPathsTest, BuildPathJoinMode)
...
+        cgroups2::container(ROOT, child1));
+    EXPECT_EQ(
+        path::join(ROOT, parent.value(), "mesos", child1.value(), "leaf"),
+        cgroups2::container(ROOT, child1, true));
+
+    EXPECT_EQ(
+        path::join(
+            ROOT,
+            parent.value(), "mesos", child1.value(), "mesos", child2.value()),
+        cgroups2::container(ROOT, child2));
```
bmahler, line 134:
```
  EXPECT_EQ("mesos/parent/mesos/child1/mesos/child2",
            cgroups2::container("mesos", child2));
```

PR #538, opened 2024-03-27
[cgroups2] Device controller tests.

DevinLeamy:
Added tests for the programs created by the the Device Controller.

### DIFF
--- a/src/tests/containerizer/cgroups2_tests.cpp
+++ b/src/tests/containerizer/cgroups2_tests.cpp
@@ -16,22 +16,32 @@
 
 #include <set>
 #include <string>
+#include <tuple>
+#include <utility>
 #include <vector>
 
 #include <process/reap.hpp>
+#include <process/gmock.hpp>
 #include <process/gtest.hpp>
 
 #include <stout/exit.hpp>
 #include <stout/foreach.hpp>
+#include <stout/gtest.hpp>
+#include <stout/os.hpp>
 #include <stout/set.hpp>
 #include <stout/tests/utils.hpp>
 #include <stout/try.hpp>
 
 #include "linux/cgroups2.hpp"
+#include "linux/ebpf.hpp"
 
+using std::pair;
 using std::set;
 using std::string;
+using std::tuple;
 using std::vector;
+
+namespace devices = cgroups2::devices;
 
 namespace mesos {
 namespace internal {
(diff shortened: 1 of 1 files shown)


src/tests/containerizer/cgroups2_tests.cpp
```diff
@@ -97,6 +111,92 @@ class Cgroups2Test : public TemporaryDirectoryTest
...
+// Combination of a path and access flags.
+typedef pair<string, int> Access;
+
+
+Try<Nothing> deviceControllerAccesses(
+    const string& cgroup,
+    const vector<devices::Entry>& allow,
+    const vector<devices::Entry>& deny,
+    const vector<Access>& allowedAccesses,
+    const vector<Access>& blockedAccesses)
```
bmahler, line 123: Let's use a parameterized test instead?

src/tests/containerizer/cgroups2_tests.cpp
```diff
@@ -210,6 +221,119 @@ TEST_F(Cgroups2Test, ROOT_CGROUPS2_EnableAndDisable)
...
+            vector<devices::Entry>{*devices::Entry::parse("c *:* rwm")},
+            vector<Access>{},
+            vector<Access>{{os::DEV_NULL, O_RDWR}}
+        },
+        DeviceControllerTestParams{
+            // allow /dev/null
+            vector<devices::Entry>{*devices::Entry::parse("c 1:3 rwm")}, 
+            vector<devices::Entry>{},
+            // read-write allowed
+            vector<Access>{{os::DEV_NULL, O_RDWR}},
```
bmahler, line 308: just fyi there's several lines here where your editor left in trailing whitespace, doesn't highlight red on github like it does on reviewboard

src/tests/containerizer/cgroups2_tests.cpp
```diff
@@ -210,6 +221,119 @@ TEST_F(Cgroups2Test, ROOT_CGROUPS2_EnableAndDisable)
...
+        DeviceControllerTestParams{
+            // allow /dev/null
+            vector<devices::Entry>{*devices::Entry::parse("c 1:3 r")}, 
+            vector<devices::Entry>{},
+            // read-only allowed
+            vector<Access>{{os::DEV_NULL, O_RDONLY}},                  
+            // read-write is blocked
+            vector<Access>{{os::DEV_NULL, O_RDWR}}                     
+        },
+        DeviceControllerTestParams{
```
bmahler, line 320: what's this one? This?

```
// Allow write-only access to /dev/null using any device type.
```

src/tests/containerizer/cgroups2_tests.cpp
```diff
@@ -210,6 +221,119 @@ TEST_F(Cgroups2Test, ROOT_CGROUPS2_EnableAndDisable)
...
+            // read is blocked
+            vector<Access>{{os::DEV_NULL, O_RDWR}, {os::DEV_NULL, O_RDONLY}} 
+        },
+        DeviceControllerTestParams{
+            vector<devices::Entry>{*devices::Entry::parse("b 1:3 r")},
+            vector<devices::Entry>{},
+            vector<Access>{},
+            // /dev/null is blocked
+            vector<Access>{{os::DEV_NULL, O_RDWR}, {os::DEV_NULL, O_RDONLY}} 
+        }
```
bmahler, line 334: Is this what we're testing?
```
// Allow access to /dev/null but with the wrong device type, therefore access is denied.
```

src/tests/containerizer/cgroups2_tests.cpp
```diff
@@ -210,6 +221,119 @@ TEST_F(Cgroups2Test, ROOT_CGROUPS2_EnableAndDisable)
...
 }
 
+// Combination of a path and access flags.
+typedef pair<string, int> Access;
+
+using DeviceControllerTestParams = tuple<
+    vector<devices::Entry>, 
+    vector<devices::Entry>,
+    vector<Access>, 
+    vector<Access>>;
```
bmahler, line 231: hm.. it's a little weird to define another Access here when there's already a devices::Entry::Access we could be using, maybe something like this instead?

```
// Combination of a path and access flags for open()ing the device.
typedef pair<string, int> OpenTest;

using DeviceControllerTestParams = tuple<
    vector<devices::Entry>, 
    vector<devices::Entry>,
    vector<OpenTest>,  // Opens that should be allowed.
    vector<OpenTest>>; // Opens that should be denied.
```

src/tests/containerizer/cgroups2_tests.cpp
```diff
@@ -210,6 +221,119 @@ TEST_F(Cgroups2Test, ROOT_CGROUPS2_EnableAndDisable)
...
+
+
+TEST_P(DeviceControllerTestFixture, ROOT_CGROUPS2_DeviceController) {
+  const string& cgroup = TEST_CGROUP;
+
+  auto params = GetParam();
+  const vector<devices::Entry> allow = std::get<0>(params);
+  const vector<devices::Entry> deny = std::get<1>(params);
+  const vector<Access> allowedAccesses = std::get<2>(params);
+  const vector<Access> blockedAccesses = std::get<3>(params);
```
bmahler, line 245: these can be refs, since the rhs is not an rvalue reference this pattern definitely copies the vectors, we take the references off when the rhs returns an rvalue reference

src/tests/containerizer/cgroups2_tests.cpp
```diff
@@ -210,6 +221,119 @@ TEST_F(Cgroups2Test, ROOT_CGROUPS2_EnableAndDisable)
...
+
+  if (!cgroups2::exists(cgroup)) {
+    ASSERT_SOME(cgroups2::create(cgroup));
+  }
```
bmahler, line 249: This is an unnecessary guard? The fixture is supposed to ensure we enter with a clean slate?

src/tests/containerizer/cgroups2_tests.cpp
```diff
@@ -210,6 +221,119 @@ TEST_F(Cgroups2Test, ROOT_CGROUPS2_EnableAndDisable)
...
+  string path = cgroups2::path(cgroup);
+
+  ASSERT_SOME(devices::configure(cgroup, allow, deny));
+  Try<vector<uint32_t>> attached = ebpf::cgroups2::attached(path);
+  EXPECT_SOME(attached);
```
bmahler, line 254: needs to be an ASSERT_SOME or the next line can crash the code dereferencing the Try

src/tests/containerizer/cgroups2_tests.cpp
```diff
@@ -210,6 +221,119 @@ TEST_F(Cgroups2Test, ROOT_CGROUPS2_EnableAndDisable)
...
+  EXPECT_EQ(1u, attached->size());
```
bmahler, line 255: Ditto here, we don't want to do attached->at(0) if this fails because it's empty?

src/tests/containerizer/cgroups2_tests.cpp
```diff
@@ -210,6 +221,119 @@ TEST_F(Cgroups2Test, ROOT_CGROUPS2_EnableAndDisable)
...
+
+  pid_t pid = ::fork();
+  ASSERT_NE(-1, pid);
```
bmahler, line 258: ideally we'd do something that's errno aware here

src/tests/containerizer/cgroups2_tests.cpp
```diff
@@ -210,6 +221,119 @@ TEST_F(Cgroups2Test, ROOT_CGROUPS2_EnableAndDisable)
...
+
+    ASSERT_SOME(ebpf::cgroups2::detach(path, attached->at(0)));
+
+    // Check that we can do both the "allowedAccesses" and "blockedAccesses".
+    foreach(const Access& access, allowedAccesses) {
+      ASSERT_SOME(os::open(access.first, access.second));
+    }
+    foreach(const Access& access, blockedAccesses) {
+      ASSERT_SOME(os::open(access.first, access.second));
+    }
```
bmahler, line 277: We can't use ASSERT in the child process, as they won't fail the test. Looks like the strategy in cgroups_tests.cpp is to use a pipe and communicate back to the parent process to indicate whether the child process' tests succeeded.

bmahler (changes requested):

src/tests/containerizer/cgroups2_tests.cpp
DevinLeamy: `/dev/null` is a `character` device. This tests that if we allow a `block` device to be read, and then try and read from a `character` device with the same type, then that request will be blocked.

src/tests/containerizer/cgroups2_tests.cpp
DevinLeamy: Sure, yeah :+1:

DevinLeamy: Renamed it `OpenArgs`, because it's the arguments for `os::open`.

src/tests/containerizer/cgroups2_tests.cpp
bmahler: It's probably easier to just have one comment that explains what each test case here does rather than these fine grained comments, for example there's no comment on the "b 1:3 rwm" test, so it's hard for the reader to know what that case is checking (e.g. different device type so not actually /dev/null?)

src/tests/containerizer/cgroups2_tests.cpp
```diff
@@ -210,6 +220,148 @@ TEST_F(Cgroups2Test, ROOT_CGROUPS2_EnableAndDisable)
...
+    vector<devices::Entry>,
+    vector<OpenArgs>,
+    vector<OpenArgs>>;
+
+class DeviceControllerTestFixture :
+    public Cgroups2Test,
+    public ::testing::WithParamInterface<DeviceControllerTestParams> {};
+
+
+TEST_P(DeviceControllerTestFixture, ROOT_CGROUPS2_DeviceController) {
```
bmahler, line 237: nit: brace on next line

src/tests/containerizer/cgroups2_tests.cpp
```diff
@@ -210,6 +220,148 @@ TEST_F(Cgroups2Test, ROOT_CGROUPS2_EnableAndDisable)
...
+  const vector<OpenArgs>& blockedAccesses = std::get<3>(params);
+
+  ASSERT_SOME(cgroups2::create(cgroup));
+  string path = cgroups2::path(cgroup);
+
+  Try<vector<uint32_t>> attached = ebpf::cgroups2::attached(path);
+  ASSERT_SOME(attached);
+  ASSERT_EQ(0u, attached->size());
+
+  EXPECT_SOME(devices::configure(cgroup, allow, deny));
```
bmahler, line 253: ASSERT since we don't want to continue the test if this fails?

src/tests/containerizer/cgroups2_tests.cpp
```diff
@@ -210,6 +220,148 @@ TEST_F(Cgroups2Test, ROOT_CGROUPS2_EnableAndDisable)
...
+      DeviceControllerTestParams{
+            vector<devices::Entry>{},
+            vector<devices::Entry>{},
+            vector<OpenArgs>{},
+            // Block accesses by default.
+            vector<OpenArgs>{{os::DEV_NULL, O_RDWR}, {os::DEV_NULL, O_RDWR}}
+       },
+       DeviceControllerTestParams{
+            vector<devices::Entry>{},
+            vector<devices::Entry>{*devices::Entry::parse("c 1:3 rwm")},
```
bmahler, line 316: indentation is inconsistent here

src/tests/containerizer/cgroups2_tests.cpp
```diff
@@ -210,6 +220,148 @@ TEST_F(Cgroups2Test, ROOT_CGROUPS2_EnableAndDisable)
...
+            // write-only allowed
+            vector<OpenArgs>{{os::DEV_NULL, O_WRONLY}},
+            // read is blocked
+            vector<OpenArgs>{{os::DEV_NULL, O_RDWR}, {os::DEV_NULL, O_RDONLY}}
+        },
+        DeviceControllerTestParams{
+            vector<devices::Entry>{*devices::Entry::parse("b 1:3 rwm")},
+            vector<devices::Entry>{},
+            vector<OpenArgs>{},
+            // /dev/null is blocked
```
bmahler, line 349: only because the entry has the wrong device type though, which isn't very clear here

bmahler (commented):
Nice, the test looks clean! We should really have some tests that actually use multiple device rules though?

PR #539, opened 2024-03-27
[cgroups2] Conditionally initialize cgroups v2 and the cgroups2 isolator.

DevinLeamy:
Based on whether the flag `--enable-cgroups-v2` is provided, on Linux machines, cgroups v2 will be used, instead of cgroups v1.

Here we initialize cgroups v2 cgroups for Mesos, similar to `assignCgroups` for cgroups v1.

If the `cgroups2/all` option is provided in the `isolation` flag, the `Cgroups2IsolatorProcess` is initialized, for resource isolation in the Mesos containerizer.

### DIFF
--- a/src/slave/containerizer/mesos/containerizer.cpp
+++ b/src/slave/containerizer/mesos/containerizer.cpp
@@ -116,6 +116,13 @@
 #include "slave/containerizer/mesos/isolators/volume/image.hpp"
 #include "slave/containerizer/mesos/isolators/volume/secret.hpp"
 #include "slave/containerizer/mesos/isolators/volume/csi/isolator.hpp"
+
+#ifdef ENABLE_CGROUPS_V2
+#include "linux/cgroups2.hpp"
+
+#include "slave/containerizer/mesos/isolators/cgroups2/cgroups2.hpp"
+#endif
+
 #endif // __linux__
 
 #if ENABLE_SECCOMP_ISOLATOR
@@ -363,6 +370,24 @@ Try<MesosContainerizer*> MesosContainerizer::create(
 
   Shared<Provisioner> provisioner = _provisioner->share();
 
+#ifdef __linux__
+  // Initialize either the cgroups v2 or cgroups v1 isolator, based on what
+  // is available on the host machine.
+  auto cgroupsIsolatorSelector = [] (const Flags& flags) -> Try<Isolator*> {
+#ifdef ENABLE_CGROUPS_V2
+    Try<bool> mounted = cgroups2::mounted();
+    if (mounted.isError()) {
+      return Error(
+          "Failed to determine if the cgroup2 filesystem is mounted: "
+          + mounted.error());
+    }
+    if (*mounted) {
+      return Cgroups2IsolatorProcess::create(flags);
+    }
+#endif // ENABLE_CGROUPS_V2
+    return CgroupsIsolatorProcess::create(flags);
+  };
+#endif // __linux__
   // Built-in isolator definitions.
   //
   // The order of the entries in this table specifies the ordering of the
@@ -424,17 +449,17 @@ Try<MesosContainerizer*> MesosContainerizer::create(
 #endif // __WINDOWS__
 
 #ifdef __linux__
-    {"cgroups/all", &CgroupsIsolatorProcess::create},
-    {"cgroups/blkio", &CgroupsIsolatorProcess::create},
-    {"cgroups/cpu", &CgroupsIsolatorProcess::create},
-    {"cgroups/cpuset", &CgroupsIsolatorProcess::create},
-    {"cgroups/devices", &CgroupsIsolatorProcess::create},
-    {"cgroups/hugetlb", &CgroupsIsolatorProcess::create},
-    {"cgroups/mem", &CgroupsIsolatorProcess::create},
-    {"cgroups/net_cls", &CgroupsIsolatorProcess::create},
-    {"cgroups/net_prio", &CgroupsIsolatorProcess::create},
-    {"cgroups/perf_event", &CgroupsIsolatorProcess::create},
-    {"cgroups/pids", &CgroupsIsolatorProcess::create},
+    {"cgroups/all", cgroupsIsolatorSelector},
+    {"cgroups/blkio", cgroupsIsolatorSelector},
+    {"cgroups/cpu", cgroupsIsolatorSelector},
+    {"cgroups/cpuset", cgroupsIsolatorSelector},
+    {"cgroups/devices", cgroupsIsolatorSelector},
+    {"cgroups/hugetlb", cgroupsIsolatorSelector},
+    {"cgroups/mem", cgroupsIsolatorSelector},
+    {"cgroups/net_cls", cgroupsIsolatorSelector},
+    {"cgroups/net_prio", cgroupsIsolatorSelector},
+    {"cgroups/perf_event", cgroupsIsolatorSelector},
+    {"cgroups/pids", cgroupsIsolatorSelector},
 
     {"appc/runtime", &AppcRuntimeIsolatorProcess::create},
     {"docker/runtime", &DockerRuntimeIsolatorProcess::create},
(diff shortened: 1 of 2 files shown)


bmahler (commented):
ENABLE_CGROUPS_V2 is a compile time variable, which we've added only to unbreak the linux builds on CI, so this is going to bail on any cgroups v1 host

as discussed offline, we'll need to check which cgroup version is being used, then go down the appropriate paths, and ideally, we can preserve the isolation flags (and instead just validate that you don't try to use inappropriate ones that aren't valid for v2 or v1)

src/slave/containerizer/mesos/isolators/cgroups2/cgroups2.cpp
```diff
@@ -30,6 +32,31 @@ namespace mesos {
...
 namespace slave {
 
+namespace cgroups2_isolator {
+
+Try<bool> supported()
+{
+  #ifndef ENABLE_CGROUPS_V2
+    return false;
+  #endif
+  if (!::cgroups2::enabled()) {
```
DevinLeamy, line 42: This checks if the host supports the `cgroup2` filesystem type. The name `cgroups2::enabled()` could be better.

src/slave/containerizer/mesos/isolators/cgroups2/cgroups2.cpp
```diff
@@ -30,6 +32,31 @@ namespace mesos {
 namespace internal {
 namespace slave {
 
+namespace cgroups2_isolator {
+
+Try<bool> supported()
```
DevinLeamy, line 37: Could modify this to either:
1. Return `bool` and convert all `return Error`s into `return false`. Or,
2. Return `Try<Nothing>` and treat `Error()` as `false` and `Nothing()` as true.

This would make it easier for the caller, as they don't have to check two things (success and value). What's your take? @bmahler

src/slave/main.cpp
```diff
@@ -406,13 +541,29 @@ int main(int argc, char** argv)
...
+        break;
+      }
+    #endif // ENABLE_CGROUPS_V2
+      // Move the agent process into its own cgroup for each of the specified
+      // subsystems if necessary before the process is initialized.
+      Try<Nothing> assign = assignCgroups(flags);
+      if (assign.isError()) {
+        EXIT(EXIT_FAILURE) << assign.error();
+      }
+    } while (false);
```
DevinLeamy, line 566: The `do-while` here looks a little weird, but it's so we can `break` out of the block if cgroups v2 is supported, to avoid having to copy to cgroups v1 initialization code. There's a few things we could do:

1. Put this in a function.
2. Have the if-elses.
3. Something else.

Happy to update this if there's a nicer approach.

src/slave/containerizer/mesos/containerizer.cpp
```diff
@@ -363,6 +368,21 @@ Try<MesosContainerizer*> MesosContainerizer::create(
...
+  #ifdef ENABLE_CGROUPS_V2
+    Try<bool> supported =
+      mesos::internal::slave::cgroups2_isolator::supported();
+    if (supported.isSome() && *supported) {
+      return Cgroups2IsolatorProcess::create(flags);
+    }
+  #endif // ENABLE_CGROUPS_V2
+
+    return CgroupsIsolatorProcess::create(flags);
+  };
```
bmahler, line 384: Hm.. it seems like mounted is already sufficient? Ditto for the main code?

```suggestion
#ifdef __linux__
  // Initialize either the cgroups v2 or cgroups v1 isolator, based on what
  // is available on the host machine.
  auto cgroupsIsolatorSelector = [] (const Flags& flags) -> Try<Isolator*> {
#ifdef ENABLE_CGROUPS_V2
    Try<bool> mounted = cgroups2::mounted();
    if (mounted.isError() {
      return Error("Failed to determine if cgroups2 is mounted": + mounted.error());
    }
    if (*mounted) {
      return Cgroups2IsolatorProcess::create(flags);
    }
#endif // ENABLE_CGROUPS_V2

    return CgroupsIsolatorProcess::create(flags);
  };
```

src/slave/main.cpp
```diff
@@ -76,6 +76,13 @@
 #ifdef __linux__
 #include "linux/cgroups.hpp"
 #include "linux/systemd.hpp"
+
+#ifdef ENABLE_CGROUPS_V2
+#include "linux/cgroups2.hpp"
+#endif // ENABLE_CGROUPS_V2
+
+#include "slave/containerizer/mesos/isolators/cgroups2/cgroups2.hpp"
```
bmahler, line 84: ifdef guard this too?

src/slave/main.cpp
bmahler: yeah maybe a lambda + early return rather than a loop + break?

src/slave/main.cpp
```diff
@@ -147,6 +155,144 @@ const char* malloc_conf = "narenas:4";
 
 
 #ifdef __linux__
+
+#ifdef ENABLE_CGROUPS_V2
+// Log any processes inside of a cgroup.
+static Try<Nothing> checkForProcesses(const string& cgroup)
```
bmahler, line 161: maybe logProcessesInCgroup?

src/slave/main.cpp
```diff
@@ -147,6 +155,144 @@ const char* malloc_conf = "narenas:4";
...
+
+    LOG(INFO) << "Found process(es) in the cgroup '" << cgroup << "'. "
+              << "Consider checking the following process(es) listed in "
+              << path::join("/sys/fs/cgroup", cgroup, "cgroup.procs")
+              << ":\n" << strings::join("\n", infos);
+  }
+
+  return Nothing();
+}
+// Initialize Mesos cgroups for cgroups v2.
```
bmahler, line 191: double newline here

src/slave/main.cpp
```diff
@@ -147,6 +155,144 @@ const char* malloc_conf = "narenas:4";
...
+//
+// Ensures that cgroups v2 is available, correctly mounted, and all of the
+// requested controllers are available. If correctly setup, the requested
+// controllers are enabled in the root cgroup, and the Mesos Agent is moved
+// into its own cgroup.
+//
+// If there are any processes in the cgroups that are created, we assume there
+// was an error in cleaning up a previous run and an error is returned.
```
bmahler, line 199: looks like this comment isn't accurate? we should probably fix the other comment that went stale as well as the code evolved to only log the processes and not bail

src/slave/main.cpp
```diff
@@ -147,6 +155,144 @@ const char* malloc_conf = "narenas:4";
...
+//
+// Creates cgroups:
+// /<root>              Top-level cgroup for the Mesos agent. Has all of the
+//                      requested subsystems enabled.
```
bmahler, line 203: let's only use "controllers" in v2 code

src/slave/containerizer/mesos/containerizer.cpp
DevinLeamy: Removed this function altogether.

src/slave/containerizer/mesos/containerizer.cpp
```diff
@@ -116,6 +116,13 @@
 #include "slave/containerizer/mesos/isolators/volume/image.hpp"
 #include "slave/containerizer/mesos/isolators/volume/secret.hpp"
 #include "slave/containerizer/mesos/isolators/volume/csi/isolator.hpp"
+
+#ifdef ENABLE_CGROUPS_V2
+#include "linux/cgroups2.hpp"
+
+#include "slave/containerizer/mesos/isolators/cgroups2/cgroups2.hpp"
+#endif
```
bmahler, line 124: // ENABLE_CGROUPS_V2

src/slave/main.cpp
```diff
@@ -147,6 +154,146 @@ const char* malloc_conf = "narenas:4";
...
+
+    LOG(INFO) << "Found process(es) in the cgroup '" << cgroup << "'. "
+              << "Consider checking the following process(es) listed in "
+              << path::join("/sys/fs/cgroup", cgroup, "cgroup.procs")
+              << ":\n" << strings::join("\n", infos);
+  }
+
+  return Nothing();
+}
+// Initialize Mesos cgroups for cgroups v2.
```
bmahler, line 190: double newline here

PR #541, opened 2024-03-29
[cgroups2] Created an interface to read and write the CPU bandwidth limit.

DevinLeamy:
In cgroups v2, the CPU bandwidth and bandwidth period (duration over which the bandwidth can be spent) are set in `cpu.max`. This patch introduces an interface to read and update these values.

A `BandwidthLimit` object is introduced, which represents a snapshot of the `cpu.max` control file.

---
Note: This stands in contrast to cgroups v1 where the period and bandwidth were set in separate
control files, `cpu.cfs_period_us` and `cpu.cfs_quota_us` respectively.

### DIFF
--- a/src/linux/cgroups2.cpp
+++ b/src/linux/cgroups2.cpp
@@ -36,6 +36,7 @@ using std::ostream;
 using std::set;
 using std::string;
 using std::vector;
+using std::to_string;
 
 using mesos::internal::fs::MountTable;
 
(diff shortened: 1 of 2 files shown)


src/linux/cgroups2.cpp
```diff
@@ -536,6 +576,59 @@ Try<Stats> parse(const string& content)
...
+}
+
+
+Try<Nothing> write(const string& cgroup, const cpu::BandwidthLimit& limit)
+{
+  if (limit.limit.isNone()) {
+    return cgroups2::write(cgroup, cpu::control::MAX, "max");
+  }
+
+  if (limit.limit.isSome() && limit.period.isNone()) {
```
bmahler, line 618: we already know that limit is some here..?

bmahler (commented):
Can you follow up with tests for this?

I'll take care of the comments and land this shortly

PR #543, opened 2024-04-01
[cgroups2] 'cpu.max' parsing hotfix and introduces a test

DevinLeamy:
The 'cpu.max' file is terminated by a newline which causes an error
while parsing. Hence, we trim whitespace before we parse its contents.

A test is also introduced.


bmahler (approved):
"hotfix" generally means when you rush out a fix to production without going through the usual process, so it isn't very clear here, we can just say "fixed a bug" or something like that

PR #546, opened 2024-04-02
[cgroups2] Introduce interface for reading threads in a cgroup.

DevinLeamy:
- `cgroups2::threads` reads the threads in a cgroup into a `set`, similar to `cgroups2::processes`.

Moving a process into a cgroup moves all the threads in that process into the cgroup. A test is introduced to ensure the threads move, as expected.

### DIFF
--- a/src/tests/containerizer/cgroups2_tests.cpp
+++ b/src/tests/containerizer/cgroups2_tests.cpp
@@ -14,8 +14,10 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#include <memory>
 #include <set>
 #include <string>
+#include <thread>
 #include <tuple>
 #include <utility>
 #include <vector>
@@ -38,7 +40,9 @@
 using std::pair;
 using std::set;
 using std::string;
+using std::thread;
 using std::tuple;
+using std::unique_ptr;
 using std::vector;
 
 namespace cpu = cgroups2::cpu;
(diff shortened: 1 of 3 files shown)


src/tests/containerizer/cgroups2_tests.cpp
```diff
@@ -181,6 +194,55 @@ TEST_F(Cgroups2Test, ROOT_CGROUPS2_AssignProcesses)
 }
 
 
+TEST_F(Cgroups2Test, ROOT_CGROUPS2_Threads)
```
DevinLeamy, line 197: Based on `ROOT_CGROUPS_AssignThreads` for in `cgroups_tests.cpp`.

src/tests/containerizer/cgroups2_tests.cpp
```diff
@@ -82,6 +87,10 @@ class Cgroups2Test : public TemporaryDirectoryTest
   {
     TemporaryDirectoryTest::SetUp();
 
+    // Move ourselves into the root cgroup, in case we moved out of the
+    // root cgroup during a previous test and didn't move back.
+    ASSERT_SOME(cgroups2::assign(cgroups2::ROOT_CGROUP, ::getpid()));
```
bmahler, line 92: can we avoid assuming that we are running in the root cgroup? it doesn't seem like something we need to assume, since the mesos code in general doesn't assume this already, it'd be nice if the tests also don't assume this?

src/tests/containerizer/cgroups2_tests.cpp
```diff
@@ -181,6 +194,55 @@ TEST_F(Cgroups2Test, ROOT_CGROUPS2_AssignProcesses)
...
+  std::thread* runningThreads[NUM_THREADS];
+
+  Latch* latch = new Latch();
+
+  for (size_t i = 0; i < NUM_THREADS; ++i) {
+    runningThreads[i] = new std::thread([=]() {
+      // Wait until the main thread tells us to exit.
+      latch->await();
+    });
+  }
```
bmahler, line 209: per offline discussion, looks like this is carried over from the v1 tests, but this can leak during test failures

also, the test overall looks like it should take the same approach of forking and moving into a test cgroup?

DevinLeamy: Updated to fork. Removed the `Latch` and made the threads `unique_ptr`s to have RAII.

src/tests/containerizer/cgroups2_tests.cpp
DevinLeamy: Removed.

src/tests/containerizer/cgroups2_tests.cpp
```diff
@@ -181,6 +185,67 @@ TEST_F(Cgroups2Test, ROOT_CGROUPS2_AssignProcesses)
...
+
+  pid_t pid = ::fork();
+  ASSERT_NE(-1, pid);
+
+  if (pid == 0) {
+    vector<unique_ptr<std::thread>> runningThreads(NUM_THREADS);
+
+    for (size_t i = 0; i < NUM_THREADS; ++i) {
+      runningThreads[i] = unique_ptr<std::thread>(new std::thread([]() {
+        while (true);
```
bmahler, line 202: maybe sleep instead of spinning on the cpu?

bmahler (commented):
This pattern of forking and then checking for failure or success exit status is quite clean!

PR #547, opened 2024-04-03
[cgroups2] Adds the `CoreControllerProcess` to the `Cgroups2IsolatorProcess`

DevinLeamy:
The `CoreControllerProcess`, referred to as the "core" controller, is
implemented and enabled by default by the cgroups v2 isolator process.

We enable it by default because all cgroups in cgroups v2 have the
core control files ("cgroup.*"), which the `CoreControllerProcess` manages.

Currently, `CoreControllerProcess` reports the number of processes and
threads in a cgroup, if the `cgroups_cpu_enable_pids_and_tids_count`
flag is provided.

### DIFF
--- a/src/slave/containerizer/mesos/isolators/cgroups2/controllers/core.cpp
+++ b/src/slave/containerizer/mesos/isolators/cgroups2/controllers/core.cpp
@@ -14,4 +14,70 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#include <set>
+#include <string>
+
+#include "linux/cgroups2.hpp"
+
+#include "slave/containerizer/mesos/isolators/cgroups2/constants.hpp"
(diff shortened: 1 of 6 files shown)


src/slave/containerizer/mesos/isolators/cgroups2/controllers/core.cpp
```diff
@@ -0,0 +1,83 @@
...
+}
+
+
+process::Future<ResourceStatistics> CoreControllerProcess::usage(
+    const ContainerID& containerId,
+    const std::string& cgroup)
+{
+  ResourceStatistics stats;
+
+  if (flags.cgroups_cpu_enable_pids_and_tids_count) {
```
DevinLeamy, line 62: Could have an "if guard" here, but I opted to keep with the pattern of: define `ResourceStatistics`, add statistics, and return. Additionally makes it easier to add more statistics.

PR #549, opened 2024-04-05
[cgroups2] Introduce interface to get cgroups nested inside of a cgroup.

DevinLeamy:
Introduces
```
	cgroups2::get(cgroup)
```
which returns the cgroups inside of the given cgroup.

### DIFF
--- a/src/linux/cgroups2.cpp
+++ b/src/linux/cgroups2.cpp
@@ -14,6 +14,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#include <fts.h>
+
 #include "linux/cgroups2.hpp"
 
 #include <iterator>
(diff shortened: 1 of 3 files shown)


src/linux/cgroups2.cpp
```diff
@@ -306,6 +308,45 @@ bool exists(const string& cgroup)
...
+    fts_close(tree);
+    return error;
+  }
+
+  if (fts_close(tree) != 0) {
+    return ErrnoError("Failed to stop traversing file system");
+  }
+
+  return cgroups;
+}
```
DevinLeamy, line 347: Not reinventing the wheel here, this is largely taken from `cgroups::get`. Ideally, we could use `std::filesystem::recursive_directory_iterator`, but it's a C++17 addition and we compile with C++11.

DevinLeamy: `cgroups::get`: https://github.com/apache/mesos/blob/b23908fa9c520a941111bf4c56ceee64d14331f1/src/linux/cgroups.cpp#L770

PR #551, opened 2024-04-05
[cgroups2] Introduce utility to parse a container id from a cgroup path.

DevinLeamy:
During agent recovery, we parse the directories in the cgroup hierarchy to determine what containers were previously running in the agent.

Here we implement the cgroup directory parsing for cgroups v2's updated cgroup directory structure.

### DIFF
--- a/src/slave/containerizer/mesos/paths.hpp
+++ b/src/slave/containerizer/mesos/paths.hpp
@@ -320,6 +320,17 @@ std::string container(
     const ContainerID& containerId,
     bool leaf = false);
 
+
+// Get a containerId from a container's cgroup.
+// Inverse of `cgroups2::container(root, id)`.
+//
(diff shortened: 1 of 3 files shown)


src/slave/containerizer/mesos/paths.hpp
```diff
@@ -320,6 +320,17 @@ std::string container(
     const ContainerID& containerId,
     bool leaf = false);
 
+
+// Get a containerId from a container's cgroup.
+// Inverse of `cgroups2::container(root, id)`.
+//
+// The provided cgroup can be an absolute path or a relative path.
+// Leaf paths (which end in `/leaf`) and non-leaf paths will resolve to the
+// same container id.
```
bmahler, line 329: to be consistent with our other "cgroup" arguments and variables in the cgroups2 library, let's only support the relative path format here

PR #552, opened 2024-04-08
[cgroups2] Update the LinuxLauncher to support cgroups v2. 

DevinLeamy:
Updates the `LinuxLauncher` to use cgroups v2. Like with other
cgroup v2 functionality, the new launcher is used by default if the
host is correctly configured for cgroups v2 and Mesos has been compiled
with the --enable-cgroups-v2 flag.

### DIFF
--- a/src/slave/containerizer/mesos/linux_launcher.cpp
+++ b/src/slave/containerizer/mesos/linux_launcher.cpp
@@ -32,6 +32,9 @@
 #include <stout/stringify.hpp>
 
 #include "linux/cgroups.hpp"
+#ifdef ENABLE_CGROUPS_V2
+#include "linux/cgroups2.hpp"
+#endif // ENABLE_CGROUPS_V2
 #include "linux/ns.hpp"
 #include "linux/systemd.hpp"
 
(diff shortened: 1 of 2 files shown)


src/slave/containerizer/mesos/linux_launcher.cpp
```diff
@@ -53,17 +56,24 @@ namespace mesos {
...
 namespace slave {
 
+struct CgroupsInfo
+{
+  // Flag indicating whether cgroups v2 is used.
+  bool cgroupsV2;
+
+  Option<std::string> freezerHierarchy;
+
+  Option<std::string> systemdHierarchy;
```
bmahler, line 66: it would be good if we could explain when these are set, e.g. if cgroupsv1 then freezer is set, but when is systemd set and how should it be used?

src/slave/containerizer/mesos/linux_launcher.cpp
```diff
@@ -251,48 +303,104 @@ Future<ContainerStatus> LinuxLauncher::status(
...
+  // can have a complete list of the containers we might ever want to
+  // destroy as well as be able to determine orphans below.
+  hashset<ContainerID> expected = {};
+
+  foreach (const ContainerState& state, states) {
+    expected.insert(state.container_id());
+
+    if (!containers.contains(state.container_id())) {
+      // The fact that we did not have a freezer (or systemd) cgroup
+      // for this container implies this container has already been
```
bmahler, line 326: this looks stale now in the v2 case?

also, this logic looks wrong? the logic is now looking in containers before it actually recovered them, so it looks like it's the wrong order? did the root tests pass with this code?

bmahler (commented):
The recovery logic looks wrong here? I'm surprised the v1 tests would pass with this, did you run them? See comment below:

src/slave/containerizer/mesos/linux_launcher.cpp
DevinLeamy: You're right. I've updated the order to recover before looping here :+1:

bmahler (commented):
I'll pull out some of these cleanups and get this landed.

PR #554, opened 2024-04-15
[cgroups2] Introduced API to set the hard memory minimum for a cgroup.

DevinLeamy:
Introduces
```
  cgroups2::memory::min(cgroup)            // get the minimum
  cgroups2::memory::set_min(cgroup, bytes) // set the minimum
```
to get and set the minimum memory in bytes that are guaranteed to not be reclaimed by the kernel under any conditions.

### DIFF
--- a/src/linux/cgroups2.cpp
+++ b/src/linux/cgroups2.cpp
@@ -772,6 +772,7 @@ namespace memory {
 namespace control {
 
 const string CURRENT = "memory.current";
+const string MIN = "memory.min";
 
 } // namespace control {
 
@@ -784,6 +785,23 @@ Try<Bytes> usage(const string& cgroup)
   }
 
   return Bytes(*contents);
+}
+
+
+Try<Nothing> set_min(const string& cgroup, const Bytes& bytes)
+{
+  return cgroups2::write(cgroup, control::MIN, stringify(bytes.bytes()));
+}
+
+
+Try<Bytes> min(const string& cgroup)
+{
+  Try<string> contents = cgroups2::read<string>(cgroup, control::MIN);
+  if (contents.isError()) {
+    return Error("Failed to read 'memory.min': " + contents.error());
+  }
+
+  return Bytes::parse(strings::trim(*contents) + "B");
 }
 
 } // namespace memory {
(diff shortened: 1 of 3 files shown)


src/linux/cgroups2.cpp
```diff
@@ -786,6 +787,23 @@ Try<Bytes> usage(const string& cgroup)
   return Bytes(*contents);
 }
 
+
+Try<Nothing> set_min(const string& cgroup, const Bytes& bytes)
+{
+  return cgroups2::write(cgroup, control::MIN, stringify(bytes.bytes()));
```
bmahler, line 793: we already have a write that takes uint64_t?

src/linux/cgroups2.cpp
```diff
@@ -786,6 +787,23 @@ Try<Bytes> usage(const string& cgroup)
...
+
+
+Try<Bytes> min(const string& cgroup)
+{
+  Try<string> contents = cgroups2::read<string>(cgroup, control::MIN);
+  if (contents.isError()) {
+    return Error("Failed to read 'memory.min': " + contents.error());
+  }
+
+  return Bytes::parse(strings::trim(*contents) + "B");
```
bmahler, line 804: no need to parse with a B here, can just cgroups::read a uint64_t and directly construct Bytes with it?

src/tests/containerizer/cgroups2_tests.cpp
```diff
@@ -325,6 +326,52 @@ TEST_F(Cgroups2Test, ROOT_CGROUPS2_MemoryUsage)
...
+  EXPECT_ERROR(cgroups2::memory::set_min(cgroups2::ROOT_CGROUP, bytes));
+
+  EXPECT_SOME(cgroups2::memory::set_min(TEST_CGROUP, bytes));
+  EXPECT_SOME_EQ(bytes, cgroups2::memory::min(TEST_CGROUP));
+}
+
+
+// Check that byte amounts written to the memory controller are rounded
+// down to the nearest page size.
+TEST_F(Cgroups2Test, ROOT_CGROUPS2_MemoryBytesRounding)
```
bmahler, line 349: is the plan to add other memory setters in here, hence not calling this a Minimum test?

PR #557, opened 2024-04-15
 [cgroups2] Introduced an interface to set a hard memory limit.

DevinLeamy:
The "memory.max" control contains the hard memory limit that a cgroup
and its descendants must remain below.

We introduce `cgroups2::memory::set_max` and `cgroups2::memory::max`
to set and get this limit.

### DIFF
--- a/src/linux/cgroups2.cpp
+++ b/src/linux/cgroups2.cpp
@@ -769,9 +769,33 @@ Try<cpu::BandwidthLimit> max(const string& cgroup)
 
 namespace memory {
 
+namespace internal {
+
+// Parse a byte limit from a string.
+//
+// Format: "max" OR <u32 bytes>
+Result<Bytes> parse_bytelimit(const string& value)
+{
+  const string& trimmed = strings::trim(value);
+  if (trimmed == "max") {
+    return None();
+  }
+
+  Try<Bytes> bytes = Bytes::parse(trimmed + "B");
+  if (bytes.isError()) {
+    return Error("Invalid byte format '" + trimmed + "': " + bytes.error());
+  }
+
+  return Bytes(*bytes);
+}
+
+} // namespace internal {
+
+
 namespace control {
 
 const string CURRENT = "memory.current";
+const string MAX = "memory.max";
 const string MIN = "memory.min";
 
 } // namespace control {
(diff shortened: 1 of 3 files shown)


src/linux/cgroups2.hpp
```diff
@@ -217,9 +217,49 @@ Try<BandwidthLimit> max(const std::string& cgroup);
...
+  // Constructs a limitless byte limit.
+  ByteLimit() = default;
+
+  ByteLimit(const Bytes& limit);
+
+  bool operator==(const ByteLimit& other) const;
+
+  // Limit in bytes. None if the limit is "unlimited".
+  Option<Bytes> limit;
+};
```
bmahler, line 234: what's up with this? is this only to avoid `Try<Option<Bytes>>`? If so, that's what `Result<Bytes>` provides:

```
Try<Nothing> set_max(const std::string& cgroup, const Option<Bytes>& limit);
Result<ByteLimit> max(const std::string& cgroup);
```

DevinLeamy: Similar to `cpu::BandwidthLimit`, a `ByteLimit` can either be some byte amount or "max". The idea is to be more explicit about Option<Bytes> representing an uncapped limit. 

IIUC we would then have:
```c++
Try<Nothing> set_max(const std::string& cgroup, const Option<Bytes>& limit);
Result<Bytes> max(const std::string& cgroup); // instead of Result<ByteLimit>? 
```

bmahler: yeah, s/ByteLimit/Bytes/ in my snippet, whoops

src/linux/cgroups2.cpp
```diff
@@ -769,9 +769,33 @@ Try<cpu::BandwidthLimit> max(const string& cgroup)
...
+{
+  const string& trimmed = strings::trim(value);
+  if (trimmed == "max") {
+    return None();
+  }
+
+  Try<Bytes> bytes = Bytes::parse(trimmed + "B");
+  if (bytes.isError()) {
+    return Error("Invalid byte format '" + trimmed + "': " + bytes.error());
+  }
```
bmahler, line 787: let's use numify<uint64_t> rather than the B string parsing hack here

src/linux/cgroups2.cpp
```diff
@@ -769,9 +769,33 @@ Try<cpu::BandwidthLimit> max(const string& cgroup)
 
 namespace memory {
 
+namespace internal {
+
+// Parse a byte limit from a string.
+//
+// Format: "max" OR <u32 bytes>
```
bmahler, line 776: u64

PR #559, opened 2024-04-16
[cgroups2] Introduce API to set a soft memory limit.

DevinLeamy:
The "memory.high" control contains the soft memory limit for a cgroup and its descendants. Exceeding the limit will cause the cgroup's processes to get throttled and will put the cgroup under memory pressure.

We introduce `cgroups2::memory::set_high` and `cgroups2::memory::high` to set and get this soft memory limit.

### DIFF
--- a/src/linux/cgroups2.cpp
+++ b/src/linux/cgroups2.cpp
@@ -796,6 +796,7 @@ namespace control {
 
 const string CURRENT = "memory.current";
 const string LOW = "memory.low";
+const string HIGH = "memory.high";
 const string MAX = "memory.max";
 const string MIN = "memory.min";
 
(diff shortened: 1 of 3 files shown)


src/linux/cgroups2.cpp
```diff
@@ -795,6 +795,7 @@ Result<Bytes> parse_bytelimit(const string& value)
 namespace control {
 
 const string CURRENT = "memory.current";
+const string HIGH = "memory.high";
 const string MAX = "memory.max";
 const string MIN = "memory.min";
```
bmahler, line 800: we can add all the control functionality here, but it looks like we're only going to use memory.low and memory.max: https://facebookmicrosites.github.io/cgroup2/docs/memory-controller.html

How crun handles requests vs limits:
https://github.com/containers/crun/blob/68df0832427e692c705bd3485719174091b37cb5/crun.1.md#memory-controller

This is how mesos would do request vs limits as well for burstable containers:
https://mesos.apache.org/documentation/latest/running-workloads/#resource-requests-and-limits

src/linux/cgroups2.hpp
```diff
@@ -275,6 +275,24 @@ Try<Nothing> set_max(const std::string& cgroup, const Option<Bytes>& limit);
...
+// Cannot be used for the root cgroup.
+Try<Nothing> set_soft_max(
+    const std::string& cgroup, const Option<Bytes>& limit);
+
+
+// Get the soft memory limit for a cgroup and its descendants.
+// If the returned limit is None, then there is no soft memory limit.
+//
+// Cannot be used for the root cgroup.
+Result<Bytes> soft_max(const std::string& cgroup);
```
bmahler, line 294: let's just call this high and set_high to keep the cgroups terminology intact and avoid any confusion, we don't really want people to have to learn our own terminology here and how it maps to the cgroups terminology

it seems to be called the "memory usage throttle limit" in cgroups v2 and fb's docs FWIW

src/linux/cgroups2.hpp
```diff
@@ -275,6 +275,24 @@ Try<Nothing> set_max(const std::string& cgroup, const Option<Bytes>& limit);
 // Cannot be used for the root cgroup.
 Result<Bytes> max(const std::string& cgroup);
 
+
+// Set the soft memory limit for a cgroup and its descendants. Exceeding the
+// soft limit will cause processes in the cgroup to be throttled and put under
+// heavy memory pressure.
```
bmahler, line 281: reclaim pressure

PR #562, opened 2024-04-17
[cgroups2] Introduces API to make cgroups threaded and check their type.

DevinLeamy:
The control 'cgroup.type' contains the type of a cgroup. Cgroups are either 'threaded' or 'domain'. Threaded cgroups can contain threads in internal cgroups but can only enable the "cpu", "cpuset", "perf_event", and "pids" controllers. Domain cgroups cannot contain internal processes but can enable any controller.

In cgroups v2, container pids are to be placed exclusively in "leaf" cgroups in the cgroup hierarchy. Controllers should not be enabled in "leaf" cgroups and they should not have any child cgroups. To help enforce these constraints, we will make "leaf cgroups" threaded, hence this commit which extends our library to allow us to make a cgroup threaded.

### DIFF
--- a/src/linux/cgroups2.hpp
+++ b/src/linux/cgroups2.hpp
@@ -101,6 +101,41 @@ Try<std::set<pid_t>> threads(const std::string& cgroup);
 // Get the absolute of a cgroup. The cgroup provided should not start with '/'.
 std::string path(const std::string& cgroup);
 
+// Cgroup types.
+// "domain": Allows the cgroup to manage resources.
+//
+// "threaded": Allows threads to live in internal nodes but can only enable
+//             the "cpu", "cpuset", "perf_event", and "pids" controllers.
+namespace type {
+
+// Domain cgroup.
+const std::string DOMAIN = "domain";
+
+// Threaded cgroup.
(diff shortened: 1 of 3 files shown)


src/linux/cgroups2.hpp
```diff
@@ -101,6 +101,41 @@ Try<std::set<pid_t>> threads(const std::string& cgroup);
...
+
+// Threaded cgroup.
+const std::string THREADED = "threaded";
+
+// Domain cgroup that serves as the root of a threaded subtree.
+const std::string THREADED_DOMAIN = "domain threaded";
+
+// Cgroup in an invalid state. Can't be assigned processes or have controllers
+// enabled.
+const std::string INVALID = "domain invalid";
```
DevinLeamy, line 122: Added for completeness; `cgroup.type` only has four possible values. 

src: https://docs.kernel.org/admin-guide/cgroup-v2.html#core-interface-files

DevinLeamy:
Since we want the isolated ("containerized") process to be able to manipulate its own cgroups, if it so chooses, we're opting to no have the leaf cgroup for each container be threaded, hence removing the motivation for this patch.

PR #563, opened 2024-04-18
  [cgroups2] Introduced API to listen for OOM events.

DevinLeamy:
Introduces `cgroups2::memory::events::oom` which returns a future that resolves when the cgroup reaches its memory limit and is about to fail with an OOM.

In cgroups v1, there was a bespoke notification API. The same does not exist in cgroups v2, meaning a new mechanism is used to detect OOM events.

Cgroups v2 provides the control 'memory.events' which contains key-value pairs of events and the number of times they took place [1]. For OOMs, we look at the value of the `oom` field. In `cgroups2::memory::events::oom` we poll, every 100ms, for changes to 'memory.events' and resolve a future when `events.oom > 0`, indicating that _"the cgroup’s memory usage was reached the limit and allocation [is] about to fail."_

This will be used inside of the `MemoryControllerProcess`, an isolator what will be provided by the `Cgroups2IsolatorProcess`.

[1] https://docs.kernel.org/admin-guide/cgroup-v2.html#memory

### DIFF
--- a/src/linux/cgroups2.cpp
+++ b/src/linux/cgroups2.cpp
@@ -24,6 +24,10 @@
 #include <string>
 #include <vector>
 
+#include <process/after.hpp>
+#include <process/loop.hpp>
+#include <process/pid.hpp>
+
 #include <stout/none.hpp>
 #include <stout/numify.hpp>
 #include <stout/os.hpp>
@@ -39,6 +43,13 @@ using std::ostream;
 using std::set;
 using std::string;
 using std::vector;
+
+using process::Break;
+using process::Continue;
+using process::ControlFlow;
+using process::Failure;
+using process::Future;
+using process::loop;
 
 using mesos::internal::fs::MountTable;
 
@@ -795,12 +806,84 @@ Result<Bytes> parse_bytelimit(const string& value)
 namespace control {
 
 const string CURRENT = "memory.current";
+const string EVENTS = "memory.events";
 const string LOW = "memory.low";
 const string HIGH = "memory.high";
 const string MAX = "memory.max";
 const string MIN = "memory.min";
 
+namespace events {
+
+Try<Events> parse(const string& content)
+{
+  const vector<string> lines = strings::split(content, "\n");
+  Events events;
+
+  foreach (const string& line, lines) {
+    if (line.empty()) {
+      continue;
+    }
+
+    vector<string> tokens = strings::split(line, " ");
+    if (tokens.size() != 2) {
+      return Error("Invalid line format in 'memory.events' expected "
+                   "<key> <value> received: '" + line + "'");
+    }
+
+    const string& field = tokens[0];
+    const string& value = tokens[1];
+
+    Try<uint64_t> count = numify<uint64_t>(value);
+    if (count.isError()) {
+      return Error("Failed to parse '" + field + "': " + count.error());
+    }
+
+    if      (field == "low")            { events.low            = *count; }
+    else if (field == "high")           { events.high           = *count; }
+    else if (field == "max")            { events.max            = *count; }
+    else if (field == "oom")            { events.oom            = *count; }
+    else if (field == "oom_kill")       { events.oom_kill       = *count; }
+    else if (field == "oom_group_kill") { events.oom_group_kill = *count; }
+  }
+
+  return events;
+}
+
+} // namespace stat {
+
 } // namespace control {
+
+namespace event {
+
+Future<Nothing> oom(const string& cgroup)
+{
+  // Watch for changes to the 'memory.events' control file. Resolve the future
+  // when the number of OOM events is non-zero.
+
+  // TODO(dleamy): Update this to use a better file-watching mechanism, rather
+  //               than polling.
+  return loop(
+    []() { return process::after(Milliseconds(100)); },
+    [cgroup](const Nothing&) -> Future<ControlFlow<Nothing>> {
+      Try<string> content =
+        cgroups2::read<string>(cgroup, memory::control::EVENTS);
+      if (content.isError()) {
+        return Failure("Failed to read 'memory.events': " + content.error());
+      }
+
+      Try<Events> events = memory::control::events::parse(strings::trim(*content));
+      if (events.isError()) {
+        return Failure("Failed to parse 'memory.events': " + events.error());
+      }
+
+      if (events->oom > 0) {
+        return Break(Nothing());
+      }
+      return Continue();
+    });
+}
+
+} // namespace event {
 
 Try<Bytes> usage(const string& cgroup)
 {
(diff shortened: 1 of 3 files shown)


bmahler (commented):
Applying these changes locally:

```cpp
diff --git a/src/linux/cgroups2.cpp b/src/linux/cgroups2.cpp
index 4271a0087..2dd197c4c 100644
--- a/src/linux/cgroups2.cpp
+++ b/src/linux/cgroups2.cpp
@@ -812,14 +812,15 @@ const string HIGH = "memory.high";
 const string MAX = "memory.max";
 const string MIN = "memory.min";
 
+} // namespace control {
+
 namespace events {
 
 Try<Events> parse(const string& content)
 {
-  const vector<string> lines = strings::split(content, "\n");
   Events events;
 
-  foreach (const string& line, lines) {
+  foreach (const string& line, strings::split(content, "\n")) {
     if (line.empty()) {
       continue;
     }
@@ -835,7 +836,7 @@ Try<Events> parse(const string& content)
 
     Try<uint64_t> count = numify<uint64_t>(value);
     if (count.isError()) {
-      return Error("Failed to parse '" + field + "': " + count.error());
+      return Error("Failed to numify '" + value + "': " + count.error());
     }
 
     if      (field == "low")            { events.low            = *count; }
@@ -849,41 +850,34 @@ Try<Events> parse(const string& content)
   return events;
 }
 
-} // namespace stat {
-
-} // namespace control {
+} // namespace events {
 
-namespace event {
 
 Future<Nothing> oom(const string& cgroup)
 {
-  // Watch for changes to the 'memory.events' control file. Resolve the future
-  // when the number of OOM events is non-zero.
-
-  // TODO(dleamy): Update this to use a better file-watching mechanism, rather
-  //               than polling.
+  // TODO(dleamy): Update this to use inotify, rather than polling.
   return loop(
-    []() { return process::after(Milliseconds(100)); },
-    [cgroup](const Nothing&) -> Future<ControlFlow<Nothing>> {
-      Try<string> content =
-        cgroups2::read<string>(cgroup, memory::control::EVENTS);
-      if (content.isError()) {
-        return Failure("Failed to read 'memory.events': " + content.error());
-      }
+      []() {
+        return process::after(Milliseconds(100));
+      },
+      [=](const Nothing&) -> Future<ControlFlow<Nothing>> {
+        Try<string> content = cgroups2::read<string>(cgroup, control::EVENTS);
+        if (content.isError()) {
+          return Failure("Failed to read 'memory.events': " + content.error());
+        }
 
-      Try<Events> events = memory::control::events::parse(strings::trim(*content));
-      if (events.isError()) {
-        return Failure("Failed to parse 'memory.events': " + events.error());
-      }
+        Try<Events> events = events::parse(strings::trim(*content));
+        if (events.isError()) {
+          return Failure("Failed to parse 'memory.events': " + events.error());
+        }
 
-      if (events->oom > 0) {
-        return Break(Nothing());
-      }
-      return Continue();
-    });
+        if (events->oom > 0) {
+          return Break(Nothing());
+        }
+        return Continue();
+      });
 }
 
-} // namespace event {
 
 Try<Bytes> usage(const string& cgroup)
 {
diff --git a/src/linux/cgroups2.hpp b/src/linux/cgroups2.hpp
index 69218e40f..91f8d65e6 100644
--- a/src/linux/cgroups2.hpp
+++ b/src/linux/cgroups2.hpp
@@ -248,42 +248,42 @@ namespace memory {
 // Snapshot of the 'memory.events' or 'memory.local.events' control files.
 struct Events
 {
-  // Number of times memory is reclaimed from the cgroup despite being under
-  // the low memory threshold. Usually indicates that the low threshold is too
-  // high.
+  // The number of times the cgroup is reclaimed due to high memory pressure
+  // even though its usage is under the low boundary. This usually indicates
+  // that the low boundary is over-committed.
   uint64_t low;
 
-  // Number of times processes inside of the cgroup were throttled and were
-  // forced to reclaim memory because the their high memory boundary was
-  // exceeded.
+  // The number of times processes of the cgroup are throttled and routed to
+  // perform direct memory reclaim because the high memory boundary was
+  // exceeded. For a cgroup whose memory usage is capped by the high limit
+  // rather than global memory pressure, this event’s occurrences are expected.
   uint64_t high;
 
-  // Number of times processes inside of a cgroup requested more memory than
-  // their hard max. This event is followed by an OOM event if the access
-  // memory cannot be reclaimed.
+  // The number of times the cgroup’s memory usage was about to go over the
+  // max boundary. If direct reclaim fails to bring it down, the cgroup goes
+  // to OOM state.
   uint64_t max;
 
-  // Number of times the cgroup reached the memory limit.
+  // The number of times the cgroup’s memory usage was reached the limit and
+  // allocation was about to fail. This event is not raised if the OOM killer
+  // is not considered as an option, e.g. for failed high-order allocations
+  // or if caller asked to not retry attempts.
   uint64_t oom;
 
-  // Number of processes inside of the cgroup subtree that were killed by any
-  // kind of OOM killer.
+  // The number of processes belonging to this cgroup killed by any kind of
+  // OOM killer.
   uint64_t oom_kill;
 
-  // Number of group OOM events, where all of the processes inside of a cgroup
-  // subtree were killed.
+  // The number of times a group OOM has occurred.
   uint64_t oom_group_kill;
 };
 
-namespace event {
 
-// Listen for an OOM event. Returns a `Future` that resolves when the cgroup
-// has reached its memory limit and is about to fail with an OOM.
+// Listen for an OOM event for the cgroup or any descendants.
 //
 // Cannot be used for the root cgroup.
 process::Future<Nothing> oom(const std::string& cgroup);
 
-} // namespace event {
 
 // Current memory usage of a cgroup and its descendants in bytes.
 Try<Bytes> usage(const std::string& cgroup);
diff --git a/src/tests/containerizer/cgroups2_tests.cpp b/src/tests/containerizer/cgroups2_tests.cpp
index 3ef208b04..c77870e05 100644
--- a/src/tests/containerizer/cgroups2_tests.cpp
+++ b/src/tests/containerizer/cgroups2_tests.cpp
@@ -505,14 +505,11 @@ TEST_F(Cgroups2Test, ROOT_CGROUPS2_OomDetection)
   const string leaf_cgroup = TEST_CGROUP + "/leaf";
   ASSERT_SOME(cgroups2::create(leaf_cgroup));
 
-  Try<os::Memory> memory = os::memory();
-  ASSERT_SOME(memory);
-
   MemoryTestHelper helper;
   ASSERT_SOME(helper.spawn());
   ASSERT_SOME(helper.pid());
 
-  Bytes limit = Megabytes(32);
+  Bytes limit = Megabytes(64);
 
   ASSERT_SOME(cgroups2::assign(leaf_cgroup, *helper.pid()));
   ASSERT_SOME(cgroups2::memory::set_max(TEST_CGROUP, limit));
```

PR #564, opened 2024-04-19
[cgroups2] Introduces an interface to read a subset of the memory usage statistics.

DevinLeamy:
Cgroups v2 exposes memory statistics through the 'memory.stat' control.

Here we introduce `cgroups2::memory::stats` to read a subset of the memory usage statistics into a new `memory::Stats` object. These statistics will be used by the `MemoryControllerProcess` to populate a `ResourceStatistics` object, like is done by the `MemorySubsystemProcess` in cgroups v1.

Additional statistics from the 'memory.stat' control can be included as they are required.

### DIFF
--- a/src/linux/cgroups2.cpp
+++ b/src/linux/cgroups2.cpp
@@ -811,6 +811,48 @@ const string LOW = "memory.low";
 const string HIGH = "memory.high";
 const string MAX = "memory.max";
 const string MIN = "memory.min";
+const string STAT = "memory.stat";
+
+namespace stat {
+
+Try<Stats> parse(const string& content)
+{
+  Stats stats;
+
+  foreach (const string& line, strings::split(content, "\n")) {
+    if (line.empty()) {
+      continue;
+    }
+
+    vector<string> tokens = strings::split(line, " ");
+    if (tokens.size() != 2) {
+      return Error("Invalid line format in 'memory.stat'; expected "
+                   "<key> <value> received: '" + line + "'");
+    }
+
+    const string& key = tokens[0];
+    const string& value = tokens[1];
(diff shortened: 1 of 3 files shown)


src/linux/cgroups2.hpp
```diff
@@ -243,6 +243,41 @@ Try<BandwidthLimit> max(const std::string& cgroup);
...
+//
+// Snapshot of the 'memory.stat' control file.
+//
+// Note:
+// We only record a subset of the memory statistics; a complete list can be
+// found in the kernel documentation.
+// https://docs.kernel.org/admin-guide/cgroup-v2.html#memory-interface-files
+struct Stats
+{
+  // Anonymous memory usage.
```
bmahler, line 256: let's just use the cgroups names and docs for these?

src/linux/cgroups2.cpp
```diff
@@ -964,6 +1005,17 @@ Result<Bytes> high(const string& cgroup)
...
 
+
+Try<Stats> stats(const string& cgroup)
+{
+  Try<string> contents = cgroups2::read<string>(cgroup, control::STAT);
+  if (contents.isError()) {
+    return Error("Failed to read 'memory.stat': " + contents.error());
+  }
+
+  return control::stat::parse(strings::trim(*contents));
```
bmahler, line 1016: hm.. the parse function should know how to parse without the trimming, was this actually needed?

src/linux/cgroups2.hpp
DevinLeamy: Was not a fan of the original names, but that certainly simplifies things :+1:

PR #565, opened 2024-04-19
[cgroups2] Error if `--cgroups_limit_swap` is used when cgroups v2 is used.

DevinLeamy:
Mesos does not support limiting swap memory when using cgroups v2. This is because the cgroups v2 API does not allow us to directly control the swap memory usage limit.  Rather, we can only control the combined swap _and_ RAM usage limit.

Therefore, when the `--cgroups_limit_swap` flag is provided and cgroups v2 is used we error during flag validation.

### DIFF
--- a/src/slave/flags.cpp
+++ b/src/slave/flags.cpp
@@ -37,6 +37,11 @@
 
 #ifdef __linux__
 #include "slave/containerizer/mesos/linux_launcher.hpp"
+
+#ifdef ENABLE_CGROUPS_V2
+#include "linux/cgroups2.hpp"
+#endif // ENABLE_CGROUPS_V2
+
 #endif // __linux__
 
 #include "slave/containerizer/mesos/provisioner/constants.hpp"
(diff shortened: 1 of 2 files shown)


src/slave/flags.cpp
```diff
@@ -652,8 +657,25 @@ mesos::internal::slave::Flags::Flags()
...
       "Cgroups feature flag to enable memory limits on both memory and\n"
-      "swap instead of just memory.\n",
-      false);
+      "swap instead of just memory. Not supported if cgroups v2 is used.\n",
+      false,
+      [](const bool& limit_swap) -> Option<Error> {
+#ifdef ENABLE_CGROUPS_V2
+        Try<bool> mounted = cgroups2::mounted();
+        if (mounted.isError()) {
+            return Error("Failed to check if cgroup2 filesystem is mounted: "
```
bmahler, line 666: nit: 2 space indent here and below

bmahler (approved):

bmahler:
> This is because the cgroups v2 API does not allow us to directly control the swap memory usage limit. Rather, we can only control the combined swap and RAM usage limit.

This doesn't sound accurate? in v2 we are able to separately control swap usage

DevinLeamy:
> > This is because the cgroups v2 API does not allow us to directly control the swap memory usage limit. Rather, we can only control the combined swap and RAM usage limit.
> 
> This doesn't sound accurate? in v2 we are able to separately control swap usage

It's the opposite - v1 combines RAM and swap and v2 gives direct control over both; you're right :)

PR #566, opened 2024-04-23
[cgroups2] Introduces API to read the peak memory usage.

DevinLeamy:
The cgroups2 control file `memory.peak` contains the maximum recorded memory usage for a cgroup and its descendants since creation.

Here we introduce `cgroups2::memory::peak_usage` to read this maximum.

---

`memory.peak` doesn’t always exist. It was introduced in May 2022 and was backported into RHEL 9.3 but our CentOS 9 distribution doesn’t use it.

src: https://stackoverflow.com/questions/66291245/is-it-possible-to-implement-max-usage-in-bytes-for-cgroup-v2

### DIFF
--- a/src/linux/cgroups2.cpp
+++ b/src/linux/cgroups2.cpp
@@ -811,6 +811,7 @@ const string LOW = "memory.low";
 const string HIGH = "memory.high";
 const string MAX = "memory.max";
 const string MIN = "memory.min";
+const string PEAK = "memory.peak";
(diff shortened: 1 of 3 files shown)


DevinLeamy:
Was intended to be used when there is an OOM event to report the maximum memory usage. Opted to use the value of `memory.current` at the time when the OOM takes place instead, because `memory.peak` does not exist on all hosts that support cgroups v2.

PR #567, opened 2024-04-23
[cgroups2] Update Agent test fixture to work on both cgroups v1 and v2 machines.

DevinLeamy:
`StartSlave()` and similar test-setup functions mounted cgroups v1 hierarchies and initialized controllers. On cgroups v2 machines, this setup would fail or result in irregular cgroup setups. As a step towards end-to-end testing for the MesosContainerizer, we update the Agent test fixtures such that they work correctly on both cgroups v1 and v2 hosts.

### DIFF
--- a/src/tests/mesos.cpp
+++ b/src/tests/mesos.cpp
@@ -36,6 +36,11 @@
 #ifdef __linux__
 #include "linux/cgroups.hpp"
 #include "linux/fs.hpp"
+
+#ifdef ENABLE_CGROUPS_V2
+#include "linux/cgroups2.hpp"
+#endif // ENABLE_CGROUPS_V2
+
 #endif
 
 #ifdef ENABLE_PORT_MAPPING_ISOLATOR
@@ -55,6 +60,7 @@
 using mesos::master::detector::MasterDetector;
 
 using std::list;
+using std::set;
 using std::shared_ptr;
 using std::string;
 using std::vector;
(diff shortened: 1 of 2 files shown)


src/tests/mesos.cpp
```diff
@@ -645,9 +664,16 @@ void ContainerizerTest<slave::MesosContainerizer>::TearDownTestCase()
...
+#ifdef ENABLE_CGROUPS_V2
+    // Clean up test cgroups.
+    if (cgroups2::exists(TEST_CGROUPS_ROOT)) {
+      ASSERT_SOME(cgroups2::destroy(TEST_CGROUPS_ROOT));
+    }
+#endif // ENABLE_CGROUPS_V2
+  } else if (cgroups::enabled() && *user == "root") {
     // Clean up any testing hierarchies.
-    Try<std::set<string>> hierarchies = cgroups::hierarchies();
+    Try<set<string>> hierarchies = cgroups::hierarchies();
```
DevinLeamy, line 676: There are a few places where I removed `std::`. This is one instance. I didn't feel it warranted a separate PR.

src/tests/mesos.cpp
```diff
@@ -579,7 +585,13 @@ slave::Flags ContainerizerTest<slave::MesosContainerizer>::CreateSlaveFlags()
 
   // Use cgroup isolators if they're available and we're root.
   // TODO(idownes): Refactor the cgroups/non-cgroups code.
-  if (cgroups::enabled() && user.get() == "root") {
+  if (cgroupsV2() && *user == "root") {
+    // TODO(dleamy): Add the memory isolator once it's supported by the cgroups
+    //               v2 isolator.
+    flags.isolation = "cgroups/cpu";
+    flags.cgroups_hierarchy = "/sys/fs/cgroup";
+    flags.cgroups_root = TEST_CGROUPS_ROOT;
```
DevinLeamy, line 593: We don't create a unique `flags.cgroups_root` like in cgroups v1. The source of that change [[1]](https://issues.apache.org/jira/browse/MESOS-926), [[2]](https://github.com/apache/mesos/commit/0f3f8f35a73dd5f9b35d657b3912449f570243d3#diff-a6984932a4ddcf84bf808f053133927e627d81583011efe943ba46531181a4d4) doesn't draw specific attention to the change. My view, in cgroups v2, is that we can trust our creation and destruction fixtures to create and destroy the test cgroup hierarchy on every test run, removing the need for this.

The reason I don't _just follow the convention in cgroups v1_ here is because we need to know the _exact_ name of the root test cgroup to create it, inside of `SetupCgroupsV2`, and we don't have access the flag `flags.cgroups_root` directly. We _could_ store the root cgroup inside of a member variable; I thought this was simpler.

bmahler (commented):
can you split this patch to reduce the diff noise? sadly github doesn't show when there's pure code movement vs reviewboard which does, so let's just split this into two commits to make incremental landing easier:

1. pull cgroup setup / teardown into functions to ready the code for v2 support
2. add v2 support

PR #570, opened 2024-04-24
[cgroups2] Made `cgroups2::processes` optionally recursive.

DevinLeamy:
Previously, `cgroups2::processes` could only fetch processes from inside of the provided cgroup. We can now fetch all of the processes inside of a cgroup subtree by passing an (optional) `recursive` flag.

```c++
Try<std::set<pid_t>> processes(
    const std::string& cgroup,
    bool recursive = false);
```

### DIFF
--- a/src/linux/cgroups2.cpp
+++ b/src/linux/cgroups2.cpp
@@ -462,7 +462,7 @@ Try<string> cgroup(pid_t pid)
 }
 
 
-Try<set<pid_t>> processes(const string& cgroup)
+Try<set<pid_t>> processes(const string& cgroup, bool recursive)
 {
   if (!cgroups2::exists(cgroup)) {
     return Error("Cgroup '" + cgroup + "' does not exist");
@@ -485,6 +485,27 @@ Try<set<pid_t>> processes(const string& cgroup)
     }
 
     pids.insert(*pid);
+  }
+
+  if (!recursive) {
+    return pids;
+  }
+
+  Try<set<string>> cgroups = cgroups2::get(cgroup);
+  foreach (const string& cgroup, *cgroups) {
+    Try<set<pid_t>> _pids = cgroups2::processes(cgroup);
+    if (_pids.isError() && cgroups2::exists(cgroup)) {
+      return Error(
+          "Failed to get processes in '" + cgroup + "': " + _pids.error());
+    } else if (_pids.isError()) {
+      // The cgroup was just removed. Ignore the error and treat it as if
+      // the cgroup had no processes.
+      continue;
+    }
+
+    pids.insert(
+        std::make_move_iterator(_pids->begin()),
+        std::make_move_iterator(_pids->end()));
   }
 
   return pids;
--- a/src/linux/cgroups2.hpp
+++ b/src/linux/cgroups2.hpp
@@ -93,7 +93,11 @@ Try<std::string> cgroup(pid_t pid);
 
 
 // Get the processes inside of a cgroup.
-Try<std::set<pid_t>> processes(const std::string& cgroup);
+//
+// Optionally fetch all of the processes in the cgroup subtree by setting
+// recursive=true. This will include all processes in nested cgroups.
+Try<std::set<pid_t>> processes(
+    const std::string& cgroup, bool recursive = false);
 
 
 // Get the threads inside of a cgroup.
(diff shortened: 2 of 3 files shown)


bmahler (commented):
would be good to handle the case where cgroups go away while we're walking the tree here, so something like:

```
// keep original semantics of cgroup needing to exist for top level one:
if (!cgroup::exists(cgroup)) {
  // error
}

// build set of cgroups to walk
set<string> cgroups = { cgroup };
if (recursive) {
  cgroups |= cgroups::get(cgroup); // handle error
}

// But ignore missing cgroups while we walk, due to races:
foreach (const string& cgroup, cgroups) {
  Try<string> contents = cgroups2::read<string>(cgroup, control::PROCESSES);
   if (contents.isError() && cgroup::exists(cgroup)) {
     // error
   }
}
```

src/linux/cgroups2.cpp
```diff
@@ -487,6 +487,27 @@ Try<set<pid_t>> processes(const string& cgroup)
     pids.insert(*pid);
   }
 
+  if (!recursive) {
+    return pids;
+  }
+
+  Try<set<string>> cgroups = cgroups2::get(cgroup);
+  foreach (const string& cgroup, *cgroups) {
```
bmahler, line 495: this can crash, naked de-reference without handling the error case

src/linux/cgroups2.cpp
```diff
@@ -487,6 +487,27 @@ Try<set<pid_t>> processes(const string& cgroup)
...
+    Try<set<pid_t>> _pids = cgroups2::processes(cgroup);
+    if (_pids.isError() && cgroups2::exists(cgroup)) {
+      return Error(
+          "Failed to get processes in '" + cgroup + "': " + _pids.error());
+    } else if (_pids.isError()) {
+      // The cgroup was just removed. Ignore the error and treat it as if
+      // the cgroup had no processes.
+      continue;
```
bmahler, line 503: this is going to ignore parse failures as well since you're looking at any error from the processes call, would be better to check this on read

bmahler (commented):
local edits to fix the comments:

```
diff --git a/src/linux/cgroups2.cpp b/src/linux/cgroups2.cpp
index bf44763d7..ad282695e 100644
--- a/src/linux/cgroups2.cpp
+++ b/src/linux/cgroups2.cpp
@@ -468,44 +468,40 @@ Try<set<pid_t>> processes(const string& cgroup, bool recursive)
     return Error("Cgroup '" + cgroup + "' does not exist");
   }
 
-  Try<string> contents = cgroups2::read<string>(cgroup, control::PROCESSES);
-  if (contents.isError()) {
-    return Error(
-        "Failed to read cgroup.procs in '" + cgroup + "': " + contents.error());
+  set<string> cgroups = {cgroup};
+
+  if (recursive) {
+    Try<set<string>> descendants = cgroups2::get(cgroup);
+    if (descendants.isError()) {
+      return Error("Failed to list cgroups: " + descendants.error());
+    }
+    cgroups.insert(descendants->begin(), descendants->end());
   }
 
   set<pid_t> pids;
-  foreach (const string& line, strings::split(*contents, "\n")) {
-    if (line.empty()) continue;
 
-    Try<pid_t> pid = numify<pid_t>(line);
-    if (pid.isError()) {
-      return Error(
-          "Failed to parse line '" + line + "' as a pid: " + pid.error());
+  foreach (const string& cgroup, cgroups) {
+    Try<string> contents = cgroups2::read<string>(cgroup, control::PROCESSES);
+
+    if (contents.isError() && !exists(cgroup)) {
+      continue; // Ignore missing cgroups due to races.
     }
 
-    pids.insert(*pid);
-  }
+    if (contents.isError()) {
+      return Error("Failed to read cgroup.procs in '" + cgroup + "': "
+                   + contents.error());
+    }
 
-  if (!recursive) {
-    return pids;
-  }
+    foreach (const string& line, strings::split(*contents, "\n")) {
+      if (line.empty()) continue;
 
-  Try<set<string>> cgroups = cgroups2::get(cgroup);
-  foreach (const string& cgroup, *cgroups) {
-    Try<set<pid_t>> _pids = cgroups2::processes(cgroup);
-    if (_pids.isError() && cgroups2::exists(cgroup)) {
-      return Error(
-          "Failed to get processes in '" + cgroup + "': " + _pids.error());
-    } else if (_pids.isError()) {
-      // The cgroup was just removed. Ignore the error and treat it as if
-      // the cgroup had no processes.
-      continue;
-    }
+      Try<pid_t> pid = numify<pid_t>(line);
+      if (pid.isError()) {
+        return Error("Failed to parse '" + line + "' as a pid: " + pid.error());
+      }
 
-    pids.insert(
-        std::make_move_iterator(_pids->begin()),
-        std::make_move_iterator(_pids->end()));
+      pids.insert(*pid);
+    }
   }
 
   return pids;
```

PR #571, opened 2024-04-24
[cgroups2] Update destroy to be async more robust.

DevinLeamy:
We were running into an inconsistent issue with `cgroups2::destroy`.
`cgroups2::destroy` would fail with EBUSY when removing cgroups with `rmdir`.
The error was being caused because some processes had not been killed
when `rmdir` was called on their cgroup; a cgroup with processes
cannot be destroyed.

After signalling a kill (by writing "1" to 'cgroup.kill') sometimes
processes were staying alive long enough to cause `rmdir` to fail.

Hence, we update `cgroups2::destroy` to wait after signalling a
SIGKILL for all the processes to drop before attempting to remove
the cgroups.

Since we wait a maximum of half a second, we don't want to block the
caller. Thus, we update `destroy` to be async.

### DIFF
--- a/src/linux/cgroups2.cpp
+++ b/src/linux/cgroups2.cpp
@@ -28,6 +28,7 @@
 #include <process/loop.hpp>
 #include <process/pid.hpp>
 
+#include <stout/adaptor.hpp>
 #include <stout/none.hpp>
 #include <stout/numify.hpp>
 #include <stout/os.hpp>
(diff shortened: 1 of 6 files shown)


src/slave/containerizer/mesos/isolators/cgroups2/cgroups2.hpp
```diff
@@ -149,7 +149,7 @@ class Cgroups2IsolatorProcess : public MesosIsolatorProcess
 
   process::Future<Nothing> __cleanup(
       const ContainerID& containerId,
-      const std::vector<process::Future<Nothing>>& futures);
+      const process::Future<Nothing>& future);
```
DevinLeamy, line 152: There was already a `__cleanup` function (that wasn't being used) inside of the `Cgroups2IsolatorProcess`. Therefore, rather than having to _add_ a `__cleanup` function, we just needed to update the parameters it took.

src/linux/cgroups2.cpp
```diff
@@ -381,41 +381,66 @@ Try<Nothing> kill(const std::string& cgroup)
...
   }
 
-  Try<set<string>> cgroups = cgroups2::get(cgroup);
-  if (cgroups.isError()) {
-    return Error("Failed to get nested cgroups: " + cgroups.error());
-  }
+  // Wait until all of the processes have been killed.
+  int retries = 50;
+  Future<Nothing> processes = loop(
+    []() { return process::after(Milliseconds(10)); },
```
bmahler, line 402: can you reduce this to 1 millisecond?

src/linux/cgroups2.cpp
```diff
@@ -381,41 +381,66 @@ Try<Nothing> kill(const std::string& cgroup)
...
+      }
 
-  vector<string> sorted(
-      std::make_move_iterator(cgroups->begin()),
-      std::make_move_iterator(cgroups->end()));
-  sorted.push_back(cgroup);
-  std::sort(sorted.rbegin(), sorted.rend());
+      if (pids->size() == 0) {
+        return Break();
+      }
```
bmahler, line 411: `pids->empty()`

src/linux/cgroups2.cpp
```diff
@@ -381,41 +381,66 @@ Try<Nothing> kill(const std::string& cgroup)
...
-  }
+  // Wait until all of the processes have been killed.
+  int retries = 50;
+  Future<Nothing> processes = loop(
+    []() { return process::after(Milliseconds(10)); },
+    [=](const Nothing&) mutable -> Future<ControlFlow<Nothing>> {
+      Try<set<pid_t>> pids = cgroups2::processes(cgroup, true);
+      if (pids.isError()) {
+        return Failure("Failed to fetch pids in cgroup: " + pids.error());
+      }
```
bmahler, line 407: maybe this should just Break? or if this returns Failure, then we likely want an onAny rather than a .then below, which will complicate things

src/linux/cgroups2.cpp
```diff
@@ -381,41 +381,66 @@ Try<Nothing> kill(const std::string& cgroup)
...
-  foreach (const string& cgroup, sorted) {
-    const string path = cgroups2::path(cgroup);
-    Try<Nothing> rmdir = os::rmdir(path, false);
-    if (rmdir.isError()) {
-      return Error(
-          "Failed to remove directory '" + path + "': " + rmdir.error());
+      --retries;
+      if (retries == 0) {
+        return Failure("Failed to kill all processes in cgroup");
+      }
```
bmahler, line 416: ditto here, maybe this should just Break?

src/linux/cgroups2.cpp
```diff
@@ -381,41 +381,66 @@ Try<Nothing> kill(const std::string& cgroup)
...
-  return Nothing();
+    vector<string> sorted(
+        std::make_move_iterator(cgroups->begin()),
+        std::make_move_iterator(cgroups->end()));
+    sorted.push_back(cgroup);
+    std::sort(sorted.rbegin(), sorted.rend());
+
+    foreach (const string& cgroup, sorted) {
+      const string path = cgroups2::path(cgroup);
+      Try<Nothing> rmdir = os::rmdir(path, false);
```
bmahler, line 435: hm.. ideally we ignore ENOENT here, but looking at os::rmdir, it doesn't care

might be worth just using ::rmdir directly here and looking at ENOENT, since os::rmdir is mainly providing recursion support and cross OS support, which we don't care about here

src/linux/cgroups2.cpp
DevinLeamy: Opted to convert the `.then` into a `.onAny` because I think I failure here warrants on error.

src/linux/cgroups2.cpp
DevinLeamy: I believe this should trigger a failure because otherwise we'll run into an error on `rmdir` which I think is less illustrative of the problem.

src/slave/containerizer/mesos/isolators/cgroups2/cgroups2.cpp
```diff
@@ -751,13 +751,27 @@ Future<Nothing> Cgroups2IsolatorProcess::_cleanup(
...
 
-  if (cgroups2::exists(infos[containerId]->cgroup)) {
-    Try<Nothing> destroy = cgroups2::destroy(infos[containerId]->cgroup);
-    if (destroy.isError()) {
-      return Failure(
-          "Failed to destroy cgroup '" + infos[containerId]->cgroup + "': "
-          + destroy.error());
-    }
+  if (!cgroups2::exists(infos[containerId]->cgroup)) {
+    return Nothing();
```
bmahler, line 755: we're not erasing from infos in this case anymore?

PR #577, opened 2024-04-25
 [cgroups2] Introduces the MemoryControllerProcess.

DevinLeamy:
Introduces the `MemoryControllerProcess`, the cgroups v2 memory
isolator, which will be used by the `Cgroups2IsolatorProcess`.

Unlike the `MemorySubsystemProcess`, the cgroups v1 memory isolator, we:
- Don't allow limits on swap memory to be set.
- Don't report memory pressure levels (this facility is no longer part
  of the cgroups memory controller's API)

Future work may include:
- Adding support for swap memory, and
- Reporting the (now available) memory pressure stall information

---

This patch updates the `ROOT_MemUsage` so it passes on a cgroups v2 machine using the new `MemoryControllerProcess`.

### DIFF
--- a/src/slave/containerizer/mesos/isolators/cgroups2/controllers/memory.cpp
+++ b/src/slave/containerizer/mesos/isolators/cgroups2/controllers/memory.cpp
@@ -1,0 +1,17 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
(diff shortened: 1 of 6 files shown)


src/slave/containerizer/mesos/isolators/cgroups2/controllers/memory.cpp
```diff
@@ -0,0 +1,203 @@
...
+
+  if (resourceRequests.mem().isNone()) {
+    return Failure("No memory resources requested");
+  }
+
+  Bytes memory = *resourceRequests.mem();
+  Bytes softLimit = std::max(memory, MIN_MEMORY);
+
+  // Set the soft memory limit.
+  Try<Nothing> high = cgroups2::memory::set_high(cgroup, softLimit);
```
ZhouJas, line 130: This should be set_low instead for the same behavior as soft limit in v1

bmahler:
Going to land this via #581 instead since it needed a minor fix.

PR #586, opened 2024-09-25
ADD: cni support for the docker executor.

andreaspeters:
With this PR I would like to add support for CNI (ContainerNetworkInterface) to the Docker Executor. As far as possible, I have followed the CNI implementation of the Mesos Containerizer. 
The following parameters and their default values are added to the Docker Executor:

- network_cni_plugins_dir: /usr/lib/cni/
- network_cni_config_dir: /etc/mesos/cni/net.d

**How is CNI used with Docker?**

By setting a name on the mesosproto NetworkInfo object which matches the CNI name in the CNI configuration.

Here is an example of a CNI configuration:

```
cat /etc/mesos/cni/net.d/10-mesos-net.conf
{
    "cniVersion": "0.2.0",
    "name": "mesos-net",
    "type": "ipvlan",
    "bridge": "cni0",
    "isGateway": true,
    "ipMasq": true,
    "vlanId": 5,
    "ipam": {
       "type": "host-local",
       "ranges": [
          [
            {
              "subnet": "10.10.0.0/16",
              "rangeStart": "10.10.0.10",
              "rangeEnd": "10.10.0.250"
            }
          ]
        ],
        "routes": [
            { "dst": "0.0.0.0/0" }
        ]
    }
}

```

**How does the Docker Executor work?**

If the name of the mesosproto NetworkInfo object has been set, the executor checks whether a CNI configuration exists for it. If this is the case, it is used to create a network interface in the container. If no CNI configuration exists, an error message appears in the stderr file in the sandbox directory. As the name can also be a Docker Network Plugin, the container is started in any case. If there is no Docker Network Plugin with the name, the container runs without an additional network interface. 

**How can the CNI support be tested?**

I have added a test that can be used via the Mesos Test Tool.

`mesos-tests --verbose --gtest_filter="DockerCniTest.*”`

### DIFF
--- a/src/docker/executor.cpp
+++ b/src/docker/executor.cpp
@@ -15,6 +15,7 @@
 // limitations under the License.
 
 #include <stdio.h>
+#include <dirent.h>
 
 #include <map>
 #include <string>
@@ -22,6 +23,7 @@
 #include <mesos/executor.hpp>
 #include <mesos/mesos.hpp>
 
+#include <process/io.hpp>
 #include <process/collect.hpp>
 #include <process/delay.hpp>
 #include <process/id.hpp>
@@ -37,6 +39,7 @@
 #include <stout/json.hpp>
 #include <stout/lambda.hpp>
 #include <stout/os.hpp>
+#include <stout/os/which.hpp>
 #include <stout/protobuf.hpp>
 #include <stout/try.hpp>
 #ifdef __WINDOWS__
@@ -65,6 +68,9 @@
 #include "messages/messages.hpp"
 
 #include "slave/constants.hpp"
+
+namespace io = process::io;
+namespace spec = mesos::internal::slave::cni::spec;
 
 using namespace mesos;
 using namespace process;
@@ -101,7 +107,9 @@ public:
       const string& launcherDir,
       const map<string, string>& taskEnvironment,
       const Option<ContainerDNSInfo>& defaultContainerDNS,
-      bool cgroupsEnableCfs)
+      bool cgroupsEnableCfs,
+      const string& network_cni_plugins_dir,
+      const string& network_cni_config_dir)      
     : ProcessBase(ID::generate("docker-executor")),
       killed(false),
       terminated(false),
@@ -116,6 +124,8 @@ public:
       taskEnvironment(taskEnvironment),
       defaultContainerDNS(defaultContainerDNS),
       cgroupsEnableCfs(cgroupsEnableCfs),
+      network_cni_plugins_dir(network_cni_plugins_dir),
+      network_cni_config_dir(network_cni_config_dir),      
       stop(Nothing()),
       inspect(Nothing()) {}
 
@@ -285,6 +295,7 @@ public:
       inspectLoop.then(defer(self(), [=](const Docker::Container& container) {
         if (!killed) {
           containerPid = container.pid;
+          containerId = container.id;
 
           // TODO(alexr): Use `protobuf::createTaskStatus()`
           // instead of manually setting fields.
@@ -318,7 +329,26 @@ public:
               }
             };
 
-            if (container.ipAddress.isSome()) {
+            bool useCNI = false;
+#ifdef __linux__  
+            if (network_cni_config_dir.size() > 0 && network_cni_plugins_dir.size() > 0) {
+              Try<std::tuple<Subprocess, std::string>> cni = attachCNI(task);
+              if (!cni.isError()) {
+                Subprocess cniSub = std::get<0>(cni.get());
+                std::string cniNetworkName = std::get<1>(cni.get());
+            
+                WIFEXITED(cniSub.status().get().get());
+                Try<std::string> cniIP = attachCNISuccess(cniSub);
+                if (cniIP.isError()) {
+                  LOG(ERROR) << cniIP.error();
+                } else {
+                  useCNI = true;
+                  setIPAddresses(cniIP.get(), false);
+                }
+              }
+            }
+#endif
+            if (container.ipAddress.isSome() && !useCNI) {            
               setIPAddresses(container.ipAddress.get(), false);
             }
 
@@ -593,6 +623,11 @@ private:
       if (!killed) {
         killed = true;
 
+#ifdef __linux__  
+        // detach container network
+        detachCNI();
+#endif        
+
         // Send TASK_KILLING if task is not killed by completion timeout and
         // the framework can handle it.
         if (!killedByTaskCompletionTimeout &&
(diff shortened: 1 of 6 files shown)


andreaspeters:
Friendly ping :-)

src/docker/executor.cpp
```diff
@@ -318,7 +329,26 @@ class DockerExecutorProcess : public ProtobufProcess<DockerExecutorProcess>
               }
             };
 
-            if (container.ipAddress.isSome()) {
+            bool useCNI = false;
+#ifdef __linux__  
+            if (network_cni_config_dir.size() > 0 && network_cni_plugins_dir.size() > 0) {
```
bmahler: use `!x.empty()`? Or it seems like these should be `Option<string>` instead to explicitly type in the optionality and avoid empty string == unset

Also, does it make sense for only one to be set? If not, should there be some validation somewhere for this?

src/docker/executor.hpp
```diff
@@ -90,6 +90,20 @@ struct Flags : public virtual mesos::internal::logging::Flags
...
+        "will find CNI plugins under these set of directories so that\n"
+        "it can execute the plugins to add/delete container from the CNI\n"
+        "networks.");
+  
+    add(&Flags::network_cni_config_dir,
+      "network_cni_config_dir",
+      "Directory path of the CNI network configuration files. For each\n"
+      "network that containers launched in Mesos agent can connect to,\n"
+      "the operator should install a network configuration file in JSON\n"
+      "format in the specified directory.");
```
bmahler: nit: inconsistent alignment

src/tests/containerizer/docker_cni_tests.cpp
```diff
@@ -0,0 +1,326 @@
...
+
+
+        return flags;
+    }
+};
+
+// This test tests the functionality of the docker's interfaces.
+TEST_F(DockerCniTest, ROOT_DOCKER_ip_interface)
+{
+    SetUp();
```
bmahler, line 222: nit: inconsistent indentation in your patch, can you sweep it to use a 2 space indent?

src/launcher/docker_executor.cpp
```diff
@@ -148,6 +148,14 @@ int main(int argc, char** argv)
       << flags.usage("Missing required option --mapped_directory");
   }
 
+  if (flags.network_cni_plugins_dir.isNone()) {
+      flags.network_cni_plugins_dir = "/usr/lib/cni/";
+  }
+
+  if (flags.network_cni_config_dir.isNone()) {
+      flags.network_cni_config_dir = "/etc/mesos/cni/net.d";
+  }
```
bmahler: These defaults look a bit odd, as if the agent's CNI flags actually have default values only when used with the docker executor? It seems like the agent flags should be directly used and therefore the operator has to set these explicitly? Will that cause a problem for you?

Right now this is a bit odd:

1. Don't set the agent's CNI flags, this gets you these default values only for docker containerizer / executor, but no defaults for the mesos containerizer. That looks strange.
2. Set the agent's CNI flags, in which case both the docker containerizer / mesos containerizer would use the values?

src/docker/executor.cpp
```diff
@@ -318,7 +329,26 @@ class DockerExecutorProcess : public ProtobufProcess<DockerExecutorProcess>
...
+            
+                WIFEXITED(cniSub.status().get().get());
+                Try<std::string> cniIP = attachCNISuccess(cniSub);
+                if (cniIP.isError()) {
+                  LOG(ERROR) << cniIP.error();
+                } else {
+                  useCNI = true;
+                  setIPAddresses(cniIP.get(), false);
+                }
+              }
```
bmahler, line 348: hm.. this looks like a bad design? docker run has already been called, *and then* we attach CNI? This means the workload could have started and failed to perform networking operations? It seems like the correct way to do this is to perform CNI work while setting up the container, whereas this looks like a hack where it's set up *after* the container has been started already?

src/docker/executor.cpp
```diff
@@ -785,6 +825,261 @@ class DockerExecutorProcess : public ProtobufProcess<DockerExecutorProcess>
...
+  // detach CNI from the container
+  Try<std::tuple<Subprocess, std::string>> detachCNI() 
+  {
+    return commandCNI(container, "DEL");
+  }
+
+  // function to attach or detach CNI from/to the container
+  Try<std::tuple<Subprocess, std::string>> commandCNI(
+    const mesos::ContainerInfo& container, 
+    const string command)
```
bmahler, line 909: feels like these should probably just return a Future of whatever information you want? not clear what the string here would represent

bmahler (commented):
This could use a more clear overview of the approach and its limitations, shouldn't we be using docker network plugins?

https://docs.docker.com/engine/extend/plugins_network/#use-network-driver-plugins

That seems like the right way to do this? Whereas the approach taken in this patch looks problematic and looks like a hack? We're doing CNI stuff **after** we've `docker run` the container?

andreaspeters:
> This could use a more clear overview of the approach and its limitations, shouldn't we be using docker network plugins?
> 
> https://docs.docker.com/engine/extend/plugins_network/#use-network-driver-plugins
> 
> That seems like the right way to do this? Whereas the approach taken in this patch looks problematic and looks like a hack? We're doing CNI stuff **after** we've `docker run` the container?

There is no official "Docker Inc" way to support CNI. They prefer to use there own docker-network plugin. Problem with that is, the plugins are pretty old. Some of them are even not in development anymore but still on the official list. 🤷🏻‍♂️

src/launcher/docker_executor.cpp
andreaspeters: 👍 your right. I think the behavior should be like with the mesos executor. I will change that to:

- It has to be set when we want to use CNI
- No default values

src/docker/executor.cpp
andreaspeters: > !x.empty()

Oh... much better, thanks. :-)

> Also, does it make sense for only one to be set?

I don't think so.

src/docker/executor.cpp
andreaspeters: We can only add CNI if the container is already running since there is no way for us to "infiltrate" the container during the startup. 

There are up to one second where the workload could have connection issues. But the container will not fail (crash). That's a point we have to write as notice.

My personal opinion is: Docker inc. should add CNI support to there engine. But they will not and on the other hand they also do not care there network plugins. 🤷🏻‍♂️ At least with these CNI implementation we can use a wide range of new network interfaces with our docker executor and does not have to use unmaintained docker network plugins.

src/docker/executor.cpp
andreaspeters: It returns the CNI name in line 1011. We need it in line 335. But maybe I also misunderstand you. 😅

PR #587, opened 2026-02-26
fix: replace 2 bare except clauses with except Exception

haosenwang1018:
Replace bare except clauses with except Exception.


bbannier:
Apache Mesos was [moved to the attic](https://attic.apache.org/projects/mesos.html) and is not being updated anymore.

timkofu:
@haosenwang1018 Here's an active fork https://github.com/m3scluster/clusterd

